PR #57, opened 2016-07-22
tweaks from API recommendations

jameschristopher:


twiecki:
Why did so many indents change? PEP8 recommends 80 char limit. I personally think there are good cases not to do that, but some of changes here revert something that was broken up elegantly.

jameschristopher:
indentation changes are an artifact of my editor, the default was incorrectly set. can fix that easily.

twiecki:
OK, lets revert those to also make it easier to review.

jameschristopher:
what indentation convention do we want to use when defining function?

https://github.com/quantopian/pyfolio/blob/master/pyfolio/plotting.py#L573-L575

or 

https://github.com/quantopian/pyfolio/blob/master/pyfolio/plotting.py#L502-L503

what indentation convention do we want to use for calling functions? there is this kind in pyfolio:

https://github.com/quantopian/pyfolio/blob/master/pyfolio/plotting.py#L360-L365

and then there is this kind also in pyfolio

https://github.com/quantopian/pyfolio/blob/master/pyfolio/plotting.py#L418-L422

twiecki:
> I'm a big fan of aligning with the left parenthese and only having one argument per line.

That's fine with me.

twiecki:
LGTM.

PR #76, opened 2016-08-24
Post release fixes

jameschristopher:


alphalens/performance.py
```diff
@@ -39,10 +39,10 @@ def factor_information_coefficient(factor,
     forward_returns : pd.DataFrame - MultiIndex
         Forward returns in indexed by date and asset.
         Separate column for each forward return window.
-    sector_adjust : bool
-        Demean forward returns by sector before computing IC.
-    by_sector : bool
-        If True, compute daily IC separately for each sector.
+    group_adjust : bool
```
twiecki: How do I pass in a custom group?

jameschristopher: At this point we assume you've already added the group to the MultiIndex. This isn't mentioned in the docstring and should be.

twiecki:
image

Seems like there's a problem with the sector codes.

jameschristopher:
Hmmm. Interesting. It's like that on master too. It doesn't seem to be a problem when running the same code on Quantopain's iPython environment. https://www.quantopian.com/posts/alphalens-a-new-tool-for-analyzing-alpha-factors

Perhaps this is environment specific.

twiecki:
:+1:

Issue #72, opened 2016-08-20
Making forward returns "demeaning" optional


luca-s:
The forward returns used for plotting are converted to returns relative to mean. While this is a very clever choice (the code is a constant surprise of many smart ideas, well done!) it implies a long short portfolio and that is not always the case. One might be interested to see if their signal can be profitable in a long (or short) only portfolio over a particular time period. So, why not making the demeaning optional?

Alphalens is shuch a great swiss knife, so why using the blade only? :D

Many thanks!

jameschristopher:
Is there a specific function you have in mind?

luca-s:
Inside create_factor_tear_sheet the following functions plotting.plot_quantile_returns_bar, plotting.plot_quantile_returns_violin and plotting.plot_cumulative_returns_by_quantile are called with returns calculated via perf.mean_return_by_quantile.

perf.mean_return_by_quantile calls utils.demean_forward_returns (the function that actually performs the "demean") and there is no way for the user to avoid that behaviour. So they end up looking at biased returns if they intend to run a long only portfolio.

Also, the function plotting.plot_cumulative_returns is called with returns calculated via perf.factor_returns. The latter function has a concept of long/short or long only portfolio (long_short argument), but there is no option for the user that simply calls create_factor_tear_sheet to switch that option.

luca-s:
PR #77

twiecki:
Closed by #77.

PR #77, opened 2016-08-25
Making forward returns "demeaning" optional

luca-s:
The forward returns used in the variour retuns related plots are converted
to returns relative to mean (demeaned). it implies a long short portfolio
and that is not always the case. This commit make the demeaning optional

see #72

### DIFF
--- a/alphalens/performance.py
+++ b/alphalens/performance.py
@@ -261,9 +261,10 @@ def quantize_factor(factor, quantiles=5, by_group=False):
 def mean_return_by_quantile(quantized_factor,
                             forward_returns,
                             by_time=None,
-                            by_group=False):
+                            by_group=False,
+                            demeaned=True):
     """
-    Computes mean demeaned returns for factor quantiles across
+    Computes mean returns for factor quantiles across
     provided forward returns columns.
 
     Parameters
@@ -281,6 +282,8 @@ def mean_return_by_quantile(quantized_factor,
     by_group : bool
         If True, compute quantile bucket returns separately for each group.
         Returns demeaning will occur on the group level.
+    demeaned : bool
+        Compute demeaned mean returns (long short portfolio)
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


alphalens/performance.py
```diff
@@ -282,8 +285,11 @@ def mean_return_by_quantile(quantized_factor,
         Standard error of returns by specified quantile.
     """
 
-    demeaned_fr = utils.demean_forward_returns(forward_returns,
+    if demeaned	:
```
twiecki: typo

luca-s: I am sorry I don't get the typo. I believe it's correct.

twiecki: Just remove the extra spaces.

luca-s: you see, there is a reason if I wear glasses and that's not enough! ;) (actually I looked at the code on github and the web page doesn't show the white space)

twiecki:
@luca-s Looks good, you think `longshort` is the best API for that? Also, you probably need to rebase.

luca-s:
Regarding longshort any suggestions for a better name?

alphalens/performance.py
```diff
@@ -290,8 +293,11 @@ def mean_return_by_quantile(quantized_factor,
         Standard error of returns by specified quantile.
     """
 
-    demeaned_fr = utils.demean_forward_returns(forward_returns,
-                                               by_group=by_group)
+    if demeaned:
+        demeaned_fr = utils.demean_forward_returns(forward_returns,
+                                                   by_group=by_group)
+    else:
+        demeaned_fr = forward_returns.copy()
```
jameschristopher: So I think we should rename the variable `demeaned_fr` because now that demeaning is optional this could be confusing we should probably change it to something like`formatted_forward_returns` or maybe even just reassign the variable `forward_returns`. Either way the variable name as it is now is no longer accurate.

jameschristopher:
@twiecki what are your thoughts on a better name? I'm not against using the argument `long_short` I'm just thinking about where in the package we are explicitly assuming a long-short portfolio and where we aren't. What's more we need to think about the confounding effects of the market on the analysis if we are long only.

twiecki:
OK, lets stick with `long_short` for now, it's pretty descriptive. I agree that there might be more subtleties when analyzing long-only portfolios.

twiecki:
Thanks @luca-s!

Issue #70, opened 2016-08-19
Average cumulative quantile return plot


luca-s:
The choice of what forward days/period to analyse with create_factor_tear_sheet is somehow arbitrary in my opinion. It would be useful to plot the average cumulative return for each quantile, with standard deviation, over a configurable period of time (this is very similar to the plot you get when running an event study). That would help understanding the average performance of each quantile and decide what days/periods to investigate with the tear sheet.

Please see attached images for a better understanding of what I mean.

![cumret1](https://cloud.githubusercontent.com/assets/15273575/17811602/b82ae21e-6623-11e6-950d-ec23b6c883ed.png)
![cumret2](https://cloud.githubusercontent.com/assets/15273575/17811605/bc76bf50-6623-11e6-86da-f8209c301fc3.png)

luca-s:
Just to show the utility of the average quantile cumulative plot I'd like to show the results of a research I did on two stock factors. The plot can easily show if a factors has alpha...

![facotr1](https://cloud.githubusercontent.com/assets/15273575/17907563/d58eaa50-697c-11e6-9bc3-41bcd45d3db8.png)

or not...

![factor2](https://cloud.githubusercontent.com/assets/15273575/17907565/d877c09e-697c-11e6-9bfc-884bfdd004e2.png)

PR #79, opened 2016-08-30
Added average cumulative quantile return plot

luca-s:
issue #70, for your evaluation

### DIFF
--- a/alphalens/plotting.py
+++ b/alphalens/plotting.py
@@ -832,7 +832,7 @@ def plot_quantile_average_cumulative_return(quantized_factor, forward_returns, b
         ax.axvline(x=0, color='k', linestyle='--')
         ax.legend()
         ax.set(ylabel='Mean Return (bps)',
-               title="Average Quantile Cumulative returns",
-               xlabel='Days')
+               title="Average Cumulative Returns by Quantile",
+               xlabel='Periods')
 
     return ax
--- a/alphalens/plotting.py
+++ b/alphalens/plotting.py
@@ -730,3 +730,109 @@ def plot_cumulative_returns_by_quantile(quantile_returns, ax=None):
     ax.axhline(1.0, linestyle='-', color='black', lw=1)
 
     return ax
+
+
+def plot_quantile_average_cumulative_return(quantized_factor, forward_returns, by_quantile=False,
+                                            periods_before=15, periods_after=15,
+                                            std_bar=False, ax=None):
+    """
+    Plots sector-wise mean daily returns for factor quantiles 
+    across provided forward price movement columns.
+    Parameters
+    ----------
+    quantized_factor : pd.Series
+        Factor quantiles indexed by date and asset.
+    forward_returns : pd.DataFrame - MultiIndex
+        Period wise forward returns in indexed by date and asset and
+        optional a custom group.
+        Separate column for each forward return window..
+    by_quantile : boolean
+        Disaggregated figures by quantile (useful to clearly see std dev bars)
+    periods_before : int
+        How many periods before factor to plot
+    periods_after  : int
+        How many periods after factor to plot
+    std_bar : boolean
+        Plot standard deviation plot
+    ax : matplotlib.Axes, optional
+        Axes upon which to plot.
+    Returns
+    -------
+    ax : matplotlib.Axes
+    """
+
+    quantiles_list = list(quantized_factor.dropna().unique())
+    quantiles = len(quantiles_list)
+
+    returns = forward_returns.copy()
+    if 'group' in returns.index.names:
+        returns.index = returns.index.droplevel(level='group')
+    returns = returns.unstack(level=['asset'])
+
+    cumulative_returns = {}
+
+    for q in quantiles_list:
+        q_fact = quantized_factor[quantized_factor == q]
+        q_returns = utils.common_start_returns(q_fact, returns,
+                                               periods_before, periods_after)
+        q_returns = q_returns.add(1).cumprod() - 1
+
+        if periods_before > 0:
+            q_returns -= q_returns.iloc[periods_before, :]
+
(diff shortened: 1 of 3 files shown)


twiecki:
Looks great, can you post a picture of the new plot?

luca-s:
Here we are ( I've just realized I need to fix the plot label "days" -> "periods") 

index3

alphalens/plotting.py
```diff
@@ -730,3 +730,109 @@ def plot_cumulative_returns_by_quantile(quantile_returns, ax=None):
...
     return ax
+
+
+def plot_quantile_average_cumulative_return(quantized_factor, forward_returns, by_quantile=False,
+                                            periods_before=15, periods_after=15,
+                                            std_bar=False, ax=None):
+    """
+    Plots sector-wise mean daily returns for factor quantiles 
+    across provided forward price movement columns.
+    Parameters
```
twiecki: Add new line before Parameters.

alphalens/plotting.py
```diff
@@ -730,3 +730,109 @@ def plot_cumulative_returns_by_quantile(quantile_returns, ax=None):
...
+    ----------
+    quantized_factor : pd.Series
+        Factor quantiles indexed by date and asset.
+    forward_returns : pd.DataFrame - MultiIndex
+        Period wise forward returns in indexed by date and asset and
```
twiecki: indexed -> index?

alphalens/plotting.py
```diff
@@ -730,3 +730,109 @@ def plot_cumulative_returns_by_quantile(quantile_returns, ax=None):
...
+        optional a custom group.
+        Separate column for each forward return window..
```
twiecki: double `..`.

alphalens/plotting.py
```diff
@@ -730,3 +730,109 @@ def plot_cumulative_returns_by_quantile(quantile_returns, ax=None):
...
+    by_quantile : boolean
```
twiecki: please add `(optional)` to all those which are.

alphalens/plotting.py
```diff
@@ -730,3 +730,109 @@ def plot_cumulative_returns_by_quantile(quantile_returns, ax=None):
...
+            ax[i].legend()
+            i += 1
+
+    else:
+
+        if ax is None:
+            f, ax = plt.subplots(1, 1, figsize=(18, 6))
+
+        i = 0
+        for quantile, q_ret in cumulative_returns.items():
```
twiecki: `for i, (quantile, q_ret) in enumerate(cumulative_returns.items()):`

alphalens/tears.py
```diff
@@ -29,7 +29,8 @@ def create_factor_tear_sheet(factor,
                              periods=(1, 5, 10),
                              quantiles=5,
                              filter_zscore=10,
-                             groupby_labels=None):
+                             groupby_labels=None,
+                             avgretplot = (5, 15) ):
```
twiecki: `avgretplot=(5, 15)):`

alphalens/utils.py
```diff
@@ -234,3 +234,54 @@ def get_clean_factor_and_forward_returns(factor,
...
+            ending_index = min(day_zero_index + after +
+                               1, len(returns.index) - 1)
+
+            series = returns.iloc[
+                starting_index:ending_index, :].loc[:, equities]
+
+            # Reset index to have the same offset (it might raise an exception)
+            series.index = range(starting_index - day_zero_index,
+                                 ending_index - day_zero_index)
+        except:
```
twiecki: Best to specify which exception(s) you are expecting here.

twiecki:
This is a great addition.

Minor comments left in the code, in addition to these:
* In #70 you showed width of the cumulative return lines, just curious why you dropped this here?
* There's quite some complexity in the grid plot layout code. I wonder if this could be simplified by using `seaborn` `facetgrid`: https://stanford.edu/~mwaskom/software/seaborn/examples/facets_with_custom_projection.html?highlight=facetgrid

alphalens/utils.py
luca-s: I would have done it if the pandas had documented the possible exceptions. If they don't specify those in the documentation I feel they might change in the future

luca-s:
> In #70 you showed width of the cumulative return lines, just curious why you dropped this here?

I dropped the width (Confidence interval) for performance reason. In #70 screenshots I used [seaborn.tsplot](https://stanford.edu/~mwaskom/software/seaborn/generated/seaborn.tsplot.html) but it is very slow unfortunately.

> There's quite some complexity in the grid plot layout code. I wonder if this could be simplified by using seaborn facetgrid:

I used the same code as the IC and plotting by group (few lines below average cumulative plot) and I agree it is too complex for what it does.

> Any ideas for unittests?

I will add one

alphalens/utils.py
twiecki: I see, I think the more conservative choice would still be to catch whatever they currently raise and fix it if they change it. That way we're not hiding potential problems.

twiecki:
>> There's quite some complexity in the grid plot layout code. I wonder if this could be simplified by using seaborn facetgrid:

> I used the same code as the IC and plotting by group (few lines below average cumulative plot) and I agree it is too complex for what it does.

Do you want to attempt to fix that? I'm also fine leaving this for a separate PR.

alphalens/utils.py
luca-s: Sounds good. Also I will catch only the first exception because the second one was only a precaution and if you don't want to hide potential problems I will remove "security belts".

luca-s:
> Do you want to attempt to fix that? I'm also fine leaving this for a separate PR.

Yes, I'd rather deal with that in another PR where all the plots are taken care of, not only the average cumulative returns one.

twiecki:
@luca-s :+1:, thanks for the contribution.

PR #81, opened 2016-09-01
Average cumulative returns by quantile: performance improvement

luca-s:
Call utils.common_start_returns just once as it is very computing
intensive.

Split the calculation of average cumulative returns from the plotting code
so that it's possible to reuse the same calculated average returns in the two
subsequent calls to plot_quantile_average_cumulative_return


twiecki:
Seems like you need to rebase.

luca-s:
Done

twiecki:
LGTM :+1:

PR #82, opened 2016-09-01
Simplified grid plot code

luca-s:
There was a lot of copy and paste of code for dealing with grid plots and
I condensed all the boilerplate in a helper class


twiecki:
This is fantastic, thanks for the contribution.

Issue #66, opened 2016-08-19
Cumulative Return by Quantile for each day/period


luca-s:
Cumulative Return by Quantile plot

Only the 1 day forward returns graph is plotted. I expected to see plots for each day/period forward return ('days' argument passed to create_factor_tear_sheet ) .
I might have a factor that shows alpha only after X or Z days and I might not be interested to see 1 day forward returns plots, so I would pass days = (X, Z) to create_factor_tear_sheet and I would expect to see the plots for X and Z forward returns. At least it would be nice to have this feature as an option if not the default behaviour.

luca-s:
I understand that is not straightforward to decide the best way to plot the cumulative daily returns plot when the forward returns have periods greater than 1 day (as the forward returns of subsequent days overlaps), but I noticed the code saves the forward returns for any period length as **daily** returns (please see #74) so the returns are already in a format suitable to be passed to the  plotting function (plotting.plot_cumulative_returns_by_quantile). Plotting the daily returns should be a reasonable compromise.

The other, more complex, option would be to calculate the daily returns as the mean of all the overlapping returns that influence that day: for a period of N days each daily returns  would be calculated as the average of N forward returns (day 1 returns of '1 day ago' factor, day 2 returns of '2 day ago' factor, day 3 returns of '3 day ago' factor and so on until N day ago factor). But that wouldn't give us much more information. 

This also applies to  #65.

luca-s:
Actually It makes me laugh that I defined "more complex" the second option. it simply requires to call rolling_mean on the daily return dataframe.

jameschristopher:
This and #65 I think are great. It is definitely misleading to look at the cumulative returns graph and see it looking terrible even though the signal was defined for a monthly or weekly predictive timeframe. Definitely PR this if you want.

luca-s:
Great, I will provide a PR then. I just hope nobody gets frustrated to see too many plots. We will see.

jameschristopher:
https://github.com/quantopian/alphalens/pull/83

Issue #65, opened 2016-08-19
Factor Weighted Cumulative Return plot for each day/period


luca-s:
Factor Weighted Cumulative Return plot

Only the 1 day forward returns graph is plotted in  'Factor Weighted Long/Short Portfolio Cumulative Return'. I expected to see plots for each day/period  forward return ('days' argument passed to create_factor_tear_sheet) .
I might have a factor that shows alpha only after X or Y days and I might not be interested to see 1 day forward returns plots, so I would pass days = (X, Y) to create_factor_tear_sheet and I would expect to see the plots for X and Y forward returns. At least it would be nice to have this feature as an option if not the default behaviour.

jameschristopher:
https://github.com/quantopian/alphalens/pull/83

PR #83, opened 2016-09-02
Plot Cumulative Return for each period

luca-s:
Cumulative Return by Quantile plot for each period and
Factor Weighted Cumulative Return plot for each period

Issue #66 and #65

### DIFF
--- a/alphalens/plotting.py
+++ b/alphalens/plotting.py
@@ -350,7 +350,7 @@ def plot_quantile_returns_bar(mean_ret_by_q,
         if ax is None:
             v_spaces = ((num_group - 1) // 2) + 1
             f, ax = plt.subplots(v_spaces, 2, sharex=False,
-                                 sharey=True, figsize=(18, 6*v_spaces))
+                                 sharey=True, figsize=(18, 6 * v_spaces))
             ax = ax.flatten()
 
         for a, (sc, cor) in zip(ax, mean_ret_by_q.groupby(level='group')):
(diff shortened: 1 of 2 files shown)


luca-s:
untitled

alphalens/plotting.py
```diff
@@ -709,7 +721,15 @@ def plot_cumulative_returns_by_quantile(quantile_returns, ax=None):
...
 
     ret_wide = quantile_returns.reset_index()\
-        .pivot(index='date', columns='quantile', values=1)
+        .pivot(index='date', columns='quantile', values=period)
+
+    if period > 1:
+        daily_returns = lambda ret, period: (
+            np.nanmean(ret) ** (1. / period)) - 1
+        period_ret_wide = ret_wide.add(1)**period
+        ret_wide = period_ret_wide.rolling(
```
jameschristopher: Is this way of doing rolling operations pandas 16 compatible? I thought this was just recently introduced.

jameschristopher:
This is awesome

alphalens/plotting.py
luca-s: Yeah, it's a pandas 18 (18.1?) feature. I wasn't sure what was the target pandas version. I will use the old method then (and eventually we have to switch back to this version as the new one complains about deprecated pandas.rolling_apply :) )

luca-s: it should be fine now

luca-s:
That makes sense, I agree with you. I will do that then.

jameschristopher:
LGTM

twiecki:
:+1:

Issue #88, opened 2016-09-20
Avoid grouping by day when calculating quantile mean returns


luca-s:
As alphalens is data frequency agnostic, the [aggregation by day performed when calculating quantile mean returns](https://github.com/quantopian/alphalens/blob/master/alphalens/tears.py#L138 ) doesn't make sense all the time. It could produce better plots for some scenarios, depending on data frequency and data time span, but not for all the possible data given in input to alphalens.

I am suggesting to remove the aggregation by day or to make it optional and in this latter case the aggregation period should be configurable: minute, day, hour, week, whatever.

Also the feature should be applied to all the plots consistently (not only on some of them). Currently it is applied only to plotting.plot_quantile_returns_violin, plotting.plot_cumulative_returns_by_quantile and plotting.plot_mean_quantile_returns_spread_time_series.

jameschristopher:
fixed in #89. Thanks Luca!

PR #89, opened 2016-09-20
Avoid grouping by day when calculating quantile mean returns

luca-s:
Issue #88 

As alphalens is data frequency agnostic, the aggregation by day performed
when calculating quantile mean returns doesn't make sense all the time. It
could produce better plots for some scenarios, depending on data
frequency and data time span, but not for all the possible data given in
input to alphalens.

I am suggesting to remove the aggregation by day


jameschristopher:
LGTM

twiecki:
:+1:

Issue #99, opened 2016-11-08
Hard coded Y axis range in "Forward Return Information Coefficient (IC)" plot


luca-s:
[This](https://github.com/quantopian/alphalens/blob/master/alphalens/plotting.py#L206) hard coded range (-0.25, 0.25) can be annoying sometimes.

jameschristopher:
We could hard code a new range  to (-1, 1)

luca-s:
Yes or don't hard code anything and let the plot to auto adjusts the axis ranges by itself. But there might be a  reason behind those hard coded values ?

jameschristopher:
My main reasoning was so that you'd be able to visually compare between multiple graphs. If the axes are different that becomes much harder

luca-s:
That's actually a good reason. So hard coding the range (-1,1) is the way to go...or we could get the maximum Y axis range amongst all the plots and subsequently set that as Y range for all the plots.

jameschristopher:
Andrew and I had talked about doing a max/min range. Sounds good! Fixed in #100

PR #100, opened 2016-11-09
Avoid hard coding Y axis range in plots

luca-s:
issue #99 

Affected plots:
- "Forward Return Information Coefficient (IC)"
- "Top Minus Bottom Quantile Mean Return"

Get the maximum Y axis range among all the plots and subsequently set
that as Y range for all the plots, so that you'd be able to visually
compare between multiple graphs. If the axes are different that becomes
much harder


jameschristopher:
Looks good. Thank you @luca-s !

Issue #98, opened 2016-11-07
Wrong mean return by Quantile


luca-s:
Here are two graphs that I got running create_factor_tear_sheet with long_short=False (I am not yet sure if this parameter matters).

 If you look at quantile 2, the "Mean return by Factor Quantile" graph shows negative mean returns for period 10 and 20, while the "Average cumulative returns by Quantile" graph shows positive mean returns at periods 10 and 20.

I am not sure where the bug is, but it seems there is one.

![bug](https://cloud.githubusercontent.com/assets/15273575/20057022/68e623b2-a4e9-11e6-8f71-bc79043343b4.png)

luca-s:
And this is what I got with long_short=True

![bug2](https://cloud.githubusercontent.com/assets/15273575/20057222/b03afaac-a4ea-11e6-86ba-e44ba859ec70.png)

luca-s:
[This](https://github.com/quantopian/alphalens/blob/master/alphalens/performance.py#L323)  is probably wrong: we cannot average many 1 period 'compound' returns with an arithmetic mean.

[This](https://github.com/quantopian/alphalens/blob/master/alphalens/utils.py#L94) too is probably wrong  when applied to 1 period 'compound' returns.

luca-s:
After few tests I can confirm the bug is in the aforementioned lines of code.

A solution would be to avoid calculating the ["1 period compound returns"](https://github.com/quantopian/alphalens/blob/master/alphalens/utils.py#L57) in get_clean_factor_and_forward_returns so that we have carry around the actual N periods forward returns that can be used in arithmetic mean calculation all around the code. We can then calculate the compound returns only when needed (e.g. before plotting) on the final averaged return (instead of averaging the compound returns that is wrong).

If you like the idea I can provide a patch.

luca-s:
Another bug I discovered is the "Average Cumulative Returns by Quantile" plot is shifted by one period backward. I fixed that too in the patch

PR #101, opened 2016-11-09
BUG: Wrong mean return by Quantile

luca-s:
issue #98


twiecki:
thanks @luca-s!
@jameschristopher can you take a look at this?

luca-s:
Because of issue #102 and also because we cannot assume that we have 1 period forward return for every dates in the time period we are analysing (while it is a requirement to have pricing data that spans the factor analysis time period) I changed average_cumulative_return_by_quantile to use pricing data instead of forward_returns_period_1

twiecki:
@luca-s Is this ready for another round of review?

luca-s:
@twiecki Yes, it is. Sorry that I kept pushing commits on top of this PR.

alphalens/performance.py
```diff
@@ -220,7 +220,7 @@ def factor_alpha_beta(factor, forward_returns, factor_returns=None):
         t_alpha = reg_fit.tvalues[0]
         alpha, beta = reg_fit.params
 
-        alpha_beta.loc['Ann. alpha', period] = (1 + alpha) ** 252 - 1
+        alpha_beta.loc['Ann. alpha', period] = (1 + alpha) ** (252.0/period) - 1
```
twiecki: Good catch :+1:

alphalens/performance.py
```diff
@@ -453,9 +453,11 @@ def average_cumulative_return_by_quantile(quantized_factor, forward_returns,
     ----------
     quantized_factor : pd.Series
         Factor quantiles indexed by date and asset.
-    forward_returns : pd.Series - MultiIndex
-        Daily forward returns indexed by date and asset and
-        optional a custom group.
+    prices : pd.DataFrame
```
twiecki: Why this change?

alphalens/plotting.py
```diff
@@ -341,7 +341,7 @@ def plot_quantile_returns_bar(mean_ret_by_q,
     ax : matplotlib.Axes
         The axes that were plotted on.
     """
-
+
```
twiecki: Stray tab.

alphalens/performance.py
luca-s: I didn't want to rely on forward returns because of issue #102 and also because it's a bad assumption thinking we have 1 period forward return for every day (while it is a requirement to have daily pricing data that spans the factor analysis time period). So I changed average_cumulative_return_by_quantile to use pricing data instead of forward_returns_period_1

luca-s: So, I made a bad choice the first time I wrote average_cumulative_return_by_quantile and I wanted to fix it.

twiecki:
See minor comments @luca-s, otherwise this looks great. Thanks for the contribution.

alphalens/performance.py
twiecki: :+1:

twiecki:
Thanks @luca-s!

PR #103, opened 2016-11-11
ENH: added "Average Cumulative Returns by Quantile" plot by group

luca-s:
ENH: added "Average Cumulative Returns by Quantile" plot by group

download 3


jameschristopher:
so looking at the code it looks like that this will draw these plots if `avgretplot` is not none. Since we have a variable called `show_groupby_plots` shouldn't it also depend on whether that variable evaluates to `True` as well?

luca-s:
@jameschristopher, the plot drawing is done inside the "Group Specific Breakdown" code block executed only if "can_group_adjust and show_groupby_plots" evaluates True. So it should be already fine.

jameschristopher:
Oh, awesome! This looks good then. Thank you @luca-s

Issue #106, opened 2016-11-21
Make certain plots default to not display


twiecki:
Tear-sheet is a bit overloaded currently.

jameschristopher:
totally, I could also see breaking the `create_factor_tear_sheet` into 

- `create_full_tear_sheet`
- `create_summary_tear_sheet`
- `create_returns_tear_sheet`
- `create_information_tear_sheet`
- `create_turnover_tear_sheet`

twiecki:
Yes, that's probably a good idea.

jameschristopher:
fixed https://github.com/quantopian/alphalens/pull/130

PR #110, opened 2016-12-03
Create sub tearsheets

jameschristopher:
This breaks down the full tear sheet into multiple smaller ones covering: returns, information, and turnover analysis.

This PR is aimed mainly at issue #106 so Thomas your thoughts would be great!

This is a first start so things are pretty rough, and it definitely won't pass tests but I think there will be a lot of discussion so I'm not too worried.

### DIFF
--- a/alphalens/performance.py
+++ b/alphalens/performance.py
@@ -171,12 +171,12 @@ def factor_returns(factor, forward_returns, long_short=True):
     weights = factor.groupby(level=['date']).apply(to_weights, long_short)
     weighted_returns = forward_returns.multiply(weights, axis=0)
 
-    factor_returns = weighted_returns.groupby(level='date').sum()
+    long_short_adjusted_factor_returns = weighted_returns.groupby(level='date').sum()
 
-    return factor_returns
+    return long_short_adjusted_factor_returns
 
 
-def factor_alpha_beta(factor, forward_returns, factor_returns=None):
+def factor_alpha_beta(factor, forward_returns, long_short_adjusted_factor_returns=None):
     """
     Computes the alpha (excess returns), alpha t-stat (alpha significance),
     and beta (market exposure) of a factor. A regression is run with
@@ -191,7 +191,7 @@ def factor_alpha_beta(factor, forward_returns, factor_returns=None):
     forward_returns : pd.DataFrame - MultiIndex
         Period wise forward returns in indexed by date and asset.
         Separate column for each forward return window.
-    factor_returns : pd.DataFrame
+    long_short_adjusted_factor_returns : pd.DataFrame
         Period wise returns of dollar neutral portfolio weighted by factor value.
 
     Returns
@@ -200,24 +200,23 @@ def factor_alpha_beta(factor, forward_returns, factor_returns=None):
         A list containing the alpha, beta, a t-stat(alpha)
         for the given factor and forward returns.
     """
-    if factor_returns is None:
-        factor_returns = factor_returns(factor, forward_returns)
+    if long_short_adjusted_factor_returns is None:
+        long_short_adjusted_factor_returns = factor_returns(factor, forward_returns)
 
     universe_ret = (forward_returns.groupby(level='date').mean()
-                          .loc[factor_returns.index])
+                          .loc[long_short_adjusted_factor_returns.index])
 
-    if isinstance(factor_returns, pd.Series):
-        factor_returns.name = universe_ret.columns.values[0]
-        factor_returns = pd.DataFrame(factor_returns)
+    if isinstance(long_short_adjusted_factor_returns, pd.Series):
+        long_short_adjusted_factor_returns.name = universe_ret.columns.values[0]
+        long_short_adjusted_factor_returns = pd.DataFrame(long_short_adjusted_factor_returns)
 
     alpha_beta = pd.DataFrame()
-    for period in factor_returns.columns.values:
+    for period in long_short_adjusted_factor_returns.columns.values:
         x = universe_ret[period].values
-        y = factor_returns[period].values
+        y = long_short_adjusted_factor_returns[period].values
 
         x = add_constant(x)
         reg_fit = OLS(y, x).fit()
-        t_alpha = reg_fit.tvalues[0]
         alpha, beta = reg_fit.params
 
         alpha_beta.loc['Ann. alpha', period] = (1 + alpha) ** (252.0/period) - 1
@@ -246,8 +245,8 @@ def quantize_factor(factor, quantiles=5, by_group=False):
         Factor quantiles indexed by date and asset.
     """
 
-    def quantile_calc(x, quantiles):
-        return pd.qcut(x, quantiles, labels=False) + 1
+    def quantile_calc(x, _quantiles):
+        return pd.qcut(x, _quantiles, labels=False) + 1
 
     grouper = ['date', 'group'] if by_group else ['date']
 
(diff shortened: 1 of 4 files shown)


jameschristopher:
@luca-s Am I using your GridFigure class correctly, I kind of had to stumble through that one.

twiecki:
```
======================================================================
ERROR: test_factor_alpha_beta_0 (alphalens.tests.test_performance.PerformanceTestCase)
----------------------------------------------------------------------
  File "/home/travis/miniconda/envs/testenv/lib/python2.7/site-packages/nose_parameterized/parameterized.py", line 365, in standalone_func
    return func(*(a + p.args), **p.kwargs)
  File "/home/travis/build/quantopian/alphalens/alphalens/tests/test_performance.py", line 216, in test_factor_alpha_beta
    factor_returns=factor_returns)
TypeError: factor_alpha_beta() got an unexpected keyword argument 'factor_returns'
======================================================================
ERROR: test_quantize_factor_0 (alphalens.tests.test_performance.PerformanceTestCase)
----------------------------------------------------------------------
  File "/home/travis/miniconda/envs/testenv/lib/python2.7/site-packages/nose_parameterized/parameterized.py", line 365, in standalone_func
    return func(*(a + p.args), **p.kwargs)
  File "/home/travis/build/quantopian/alphalens/alphalens/tests/test_performance.py", line 142, in test_quantize_factor
    by_group=by_group)
  File "/home/travis/build/quantopian/alphalens/alphalens/performance.py", line 254, in quantize_factor
    factor_quantile = factor_percentile.apply(quantile_calc, quantiles=quantiles)
  File "/home/travis/miniconda/envs/testenv/lib/python2.7/site-packages/pandas/core/groupby.py", line 694, in apply
    return self._python_apply_general(f)
  File "/home/travis/miniconda/envs/testenv/lib/python2.7/site-packages/pandas/core/groupby.py", line 698, in _python_apply_general
    self.axis)
  File "/home/travis/miniconda/envs/testenv/lib/python2.7/site-packages/pandas/core/groupby.py", line 1601, in apply
    res = f(group)
  File "/home/travis/miniconda/envs/testenv/lib/python2.7/site-packages/pandas/core/groupby.py", line 685, in f
    return func(g, *args, **kwargs)
TypeError: quantile_calc() got an unexpected keyword argument 'quantiles'
======================================================================
ERROR: test_quantize_factor_1 (alphalens.tests.test_performance.PerformanceTestCase)
----------------------------------------------------------------------
  File "/home/travis/miniconda/envs/testenv/lib/python2.7/site-packages/nose_parameterized/parameterized.py", line 365, in standalone_func
    return func(*(a + p.args), **p.kwargs)
  File "/home/travis/build/quantopian/alphalens/alphalens/tests/test_performance.py", line 142, in test_quantize_factor
    by_group=by_group)
  File "/home/travis/build/quantopian/alphalens/alphalens/performance.py", line 254, in quantize_factor
    factor_quantile = factor_percentile.apply(quantile_calc, quantiles=quantiles)
  File "/home/travis/miniconda/envs/testenv/lib/python2.7/site-packages/pandas/core/groupby.py", line 694, in apply
    return self._python_apply_general(f)
  File "/home/travis/miniconda/envs/testenv/lib/python2.7/site-packages/pandas/core/groupby.py", line 698, in _python_apply_general
    self.axis)
  File "/home/travis/miniconda/envs/testenv/lib/python2.7/site-packages/pandas/core/groupby.py", line 1601, in apply
    res = f(group)
  File "/home/travis/miniconda/envs/testenv/lib/python2.7/site-packages/pandas/core/groupby.py", line 685, in f
    return func(g, *args, **kwargs)
TypeError: quantile_calc() got an unexpected keyword argument 'quantiles'
======================================================================
ERROR: test_quantize_factor_2 (alphalens.tests.test_performance.PerformanceTestCase)
----------------------------------------------------------------------
  File "/home/travis/miniconda/envs/testenv/lib/python2.7/site-packages/nose_parameterized/parameterized.py", line 365, in standalone_func
    return func(*(a + p.args), **p.kwargs)
  File "/home/travis/build/quantopian/alphalens/alphalens/tests/test_performance.py", line 142, in test_quantize_factor
    by_group=by_group)
  File "/home/travis/build/quantopian/alphalens/alphalens/performance.py", line 254, in quantize_factor
    factor_quantile = factor_percentile.apply(quantile_calc, quantiles=quantiles)
  File "/home/travis/miniconda/envs/testenv/lib/python2.7/site-packages/pandas/core/groupby.py", line 694, in apply
    return self._python_apply_general(f)
  File "/home/travis/miniconda/envs/testenv/lib/python2.7/site-packages/pandas/core/groupby.py", line 698, in _python_apply_general
    self.axis)
  File "/home/travis/miniconda/envs/testenv/lib/python2.7/site-packages/pandas/core/groupby.py", line 1601, in apply
    res = f(group)
  File "/home/travis/miniconda/envs/testenv/lib/python2.7/site-packages/pandas/core/groupby.py", line 685, in f
    return func(g, *args, **kwargs)
TypeError: quantile_calc() got an unexpected keyword argument 'quantiles'
----------------------------------------------------------------------
```

luca-s:
@jameschristopher the idea of GridFigure class is to define a grid at constructor time (  GridFigure(rows, cols) ) and then to request a subplot for each cell of the grid calling GridFigure.get_cell  (sequentially from the top left cell to the bottom right one) or calling GridFigure.get_row to get a subplot that spans multiple cells, a full row.

The way you used it seems fine, except the grid is too big [sometime](https://github.com/quantopian/alphalens/blob/1692172d56f9397703d7eddfc522f95d230cb4e8/alphalens/tears.py#L104).

jameschristopher:
Good to know. One of the the main issues I came across while writing this PR is the fact that the function signatures are varying. It would be nice to have a consistent function signature for tearsheets something like `my_tear_sheet(factor, forward_returns, <specifc_arg>, *args, **kwargs)` very similar to how [pyfolio tear sheets look.](https://github.com/quantopian/pyfolio/blob/master/pyfolio/tears.py) Of course this then puts the onus on the user to supply properly shifted and aligned `forward_returns`. Thoughts?

twiecki:
@jameschristopher Just making sure you saw the failing tests.

luca-s:
@jameschristopher I have been thinking about the varying function signature and I have a sort of idea. It's not perfect but it might be worth sharing it.

As forward_returns and quantiles Series must have the same index as factor, they could become factor's optional columns instead. In this way we could make all the functions signatures something like `my_tear_sheet(factor, price, <specifc_arg>, *args, **kwargs) `.  When a  function needs forward_returns or quantiles it will check their existence in factor's columns and if not present they will be calculated and added (to factor). forward_returns and quantiles columns would act as a cache and multiple calls to different tear sheet functions would make use of the cached values without additional computation. This also won't enforce any particular function call order and the details of this implementation are hidden from the user.

twiecki:
I like that idea @luca-s. However, changing passed-in data in-place seems not as ideal.

Another idea would be to unify all calls to:
`tear_sheet(factors, prices=None, fwd_returns=None)`. If prices are passed in, they are automatically converted to forward returns. The conversion could also be done via a wrapper.

luca-s:
@twiecki we can avoid changing passed-in data in-place and simply make use of `fwd_returns` and `quantiles` columns if present and calculate those otherwise without adding them to `factor`. So a function like `create_full_tear_sheet`  that calls several sub tearsheet functions would prepare a factor dataframe with the additional `fwd_returns` and `quantiles` columns, but for other simple cases any tearsheet function would still be available using factor and price only.

Your idea is fine too, but perf.quantize_factor would be called many times, as currently happens in `create_full_tear_sheet`,  unless we pass `quantized_factor` to every function together with `fwd_returns`.

twiecki:
I think we mean similar things, the main difference is you propose to append `fwd_returns` and `quantiles` to `factors`, while I would keep those separate and pass them into the sub-tear-sheets directly. As `fwd_returns` would only get calculated when they are not passed in the two proposals would have the same number of computation.

For some reason adding them to the dataframe seems a bit hacky. Then again, the two need to match up (in terms of common index) so from that perspective it makes sense to have them in the same DataFrame. And we would pass around less kwargs.

twiecki:
> we can avoid changing passed-in data in-place and simply make use of fwd_returns and quantiles columns if present and calculate those otherwise without adding them to factor.

But then how would they be found in the sub-tear-sheets if we don't add them?

luca-s:
We are indeed saying similar things, the only difference is the api we expose to the users. It's not a big deal anyway, so separating the arguments is fine with me too.

I just want to stress again that in your approach the base signature would be `tear_sheet(factors, prices=None, fwd_returns=None, **factor_quantile=None**, ....)`. The `factor_quantile` (called also `quantized_factor` in the code ) is needed to avoid recomputing factor quantization many times.  Do you agree? Or did I miss something?

twiecki:
We could add `factor_quantile` too I suppose. I get the feeling though that both are less than ideal.

How about a class-based approach where you create an object passing in all the relevant data once, we compute the required functions. E.g.:

```python
fa = alphalens.FactorAnalysis(factors, prices) # fwd_returns and factor_quantiles get created
fa.create_turnover_tearsheet()
```

luca-s:
I really like the OOP approach, but in that case we are forcing users to adopt OOP too in case they want to customize Alphalens behaviour.

Currently if they like to add/remove/customize some of the tear sheet plots they can copy/paste create_factor_tear_sheet code and modify it, as it is very linear and easy.  In case we go for the class-based API the process might become not so straightforward for an average user.

All that is ok with me, as long as we are aware of this implication.

twiecki:
@luca-s I agree that it's probably overkill for solving just that problem. An alternative is to have a `NamedTuple` that holds the various data structures. I suppose that's similar to your proposal of putting everything into a single DataFrame, but a bit more explicit. Then the calling signature would be `tear_sheet(factors=None, prices=None, precomputed_values=None)`.

I would actually be curious if the fwd_returns and factor quantization computation even takes up enough time to warrant trying to cache anything and just always recompute when required.

jameschristopher:
What if instead we just choose to unify the tearsheet API around `factor` and `forward_returns` so pretty much all tearsheet calls would look like `my_tear_sheet(factor, forward_returns, *args)` where the args are mostly those 'on/off' variables like showing certain plots, demeaning, etc.

Of course this means that there would be an extra step for the quant. They would first have to compute the forward returns, and add the quantile index level to the factor. I think we can definitely provide more helper functions in the `utils` module in this regard.

luca-s:
@jameschristopher the extra step for the quant might not be an issue if it is straightforward and well documented. Also you suggest to make the `quantile` a new factor index level instead of it being a separate `pd.Series` as it is now, is that right? But then why `forward_returns` deserve a different treatment and it is passed as a separate argument? I am not against your idea, but probably I cannot see the rationale behind it.

Looking at the different sub-tear-sheets created in this PR I can see some of them require `forward_returns`, some require the `quantized factor` and those two sets overlap but not completely. There is also one exception that requires `prices` instead. So, looking at those different needs I cannot say that `forward_returns` could easily be the common denominator to all the function calls. Instead `price` could be, as all the other information can be derived from it.

I feel I have probably missed something though, otherwise you wouldn't have suggested  `forward_returns` as common argument. What would the benefits be of unify the tearsheet API around `forward_returns` ?

twiecki:
I agree with both your points. It's not a big ask to call a function to convert the data to the correct format, especially if well documented. Question is what that returns. Here I've come around to @luca-s's initial proposal of just having multiple columns -- factor, fwd_returns, quantiles etc. The data is all indexed in the same way so it's only logical to combine it into a single dataframe rather than do joins etc. It's really the tidy-data mantra and I'm fairly convinced that's the way to go. I think @luca-s should already be on board since it's his proposal. What do you think @jameschristopher?

twiecki:
API would be `my_tear_sheet(factor_returns_quantiles, **kwargs)`, can't get any simpler than that.

jameschristopher:
For those who don't know [this article has a nice explanation of "tidy data"](http://www.jeannicholashould.com/tidy-data-in-python.html)

jameschristopher:
So the single data struture we'd centralize the tearsheets around is a large dataframe looking something like 
screen shot 2017-01-13 at 11 53 54 am

jameschristopher:
closing this PR as its morphed into PR https://github.com/quantopian/alphalens/pull/130

PR #111, opened 2016-12-05
Improved tests for perf.average_cumulative_return_by_quantile

luca-s:
Following @twiecki suggestion "prefer tests that have small and simple inputs that are so trivial that the correct answer is obvious" I created better tests for perf.average_cumulative_return_by_quantile. Now the test case has 4 assets with a fixed return rate each so that the results can be easily verified. 

I also discovered 2 bugs in perf.average_cumulative_return_by_quantile:
- Deamining in the Group Specific Breakdown was not performed
- Cannot apply demeaning on the daily returns and build the cumulative returns afterward. Instead we have to build the cumulative returns from the daily returns and then apply the demeaning (and finally averaging them). This doesn't affect the results too much, but still it is a bug.


alphalens/tears.py
```diff
@@ -303,7 +303,7 @@ def next_cell(self):
                                                                                     prices,
                                                                                     periods_before=before,
                                                                                     periods_after=after,
-                                                                                    demeaned=long_short)
+                                                                                    demeaned=True)
```
twiecki: Why this change?

luca-s: That is actually a bug fix.

As we are calling the function for each group separately (as we are in the "Group Specific Breakdown" code), we want returns demeaning to occur on the group level. Few lines above the code does the same when it forces demeaning on perf.mean_return_by_quantile.

twiecki:
:+1:

PR #112, opened 2016-12-12
Added sector neutral (actually group neutral) cumulative returns plot

luca-s:
When group information is provided to alphalens a group neutral cumulative returns plot is now displayed

### DIFF
--- a/alphalens/performance.py
+++ b/alphalens/performance.py
@@ -139,7 +139,7 @@ def mean_information_coefficient(factor,
     return ic
 
 
-def factor_returns(factor, forward_returns, long_short=True):
+def factor_returns(factor, forward_returns, long_short=True, group_neutral=False):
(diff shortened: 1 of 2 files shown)


twiecki:
Does this add new plots?

luca-s:
Yes, a plot similar to Factor Weighted Cumulative Returns plot, but the weights are calculated to be sector neutral (or group neutral).

screenshot from 2016-12-13 14-01-28

luca-s:
@twiecki I removed the plotting, so this PR simply add the api to plot a sector neutral portfolio (more generally a group neutral). It can be decided later to add this plot or not, maybe after the 'create-sub-tearsheet' branch has been merged.

twiecki:
Looks great, thanks @luca-s!

Issue #116, opened 2016-12-26
plot_mean_quantile_returns_spread_time_series seems failed with dataframe input


HereticSK:
# Problem
When I substitute my local data in `tear_sheet_walk_through.ipynb` and call `alphalens.plotting.plot_mean_quantile_returns_spread_time_series(quant_return_spread, std_err_spread)`, 

I got :
>Traceback (most recent call last):
>  File "D:\Anaconda3\lib\site-packages\alphalens\plotting.py", line 494, in plot_mean_quantile_returns_spread_time_series
>    ax=a)
> AttributeError: 'NoneType' object has no attribute 'get_ylim'

`qutant_return_spread` and `std_err_spread` is calculated with my local data, which look like:

![image](https://cloud.githubusercontent.com/assets/17619612/21482452/2a9a528a-cbaf-11e6-8aca-874306fb3fd6.png)
![image](https://cloud.githubusercontent.com/assets/17619612/21482454/3539a8e4-cbaf-11e6-8d96-f50b9e62bd48.png)

# Some thoughts
I think the problem might be caused by the first if block in the definition of `plot_mean_quantile_returns_spread_time_series`. 
```python

    if isinstance(mean_returns_spread, pd.DataFrame):
        if ax is None:
            ax = [None for a in mean_returns_spread.columns]

        ymin, ymax = (None, None)
        for a, (name, fr_column) in zip(ax, mean_returns_spread.iteritems()):
            stdn = None if std_err is None else std_err[name]
            a = plot_mean_quantile_returns_spread_time_series(fr_column,
                                                                   std_err=stdn,
                                                                   ax=a)
            curr_ymin, curr_ymax = a.get_ylim()
            ymin = curr_ymin if ymin is None else min(ymin, curr_ymin)
            ymax = curr_ymax if ymax is None else max(ymax, curr_ymax)                                                          

        for a in ax:
            a.set_ylim([ymin, ymax])

        return ax
```

When passed a dataframe and `ax` is `None`, the code sets `ax` to be a list of `None`s. In my example, it is `[None, None, None]`. And then it calls itself on each column of the dataframe, with key argument `ax=a`. 
```python
a = plot_mean_quantile_returns_spread_time_series(fr_column, std_err=stdn, ax=a)
```

Here, `a` is still `None`, so when it comes to `curr_ymin, curr_ymax = a.get_ylim()`, it raises an error 
> AttributeError: 'NoneType' object has no attribute 'get_ylim'

[...]

PR #117, opened 2016-12-29
Fix issue #116 plot_mean_quantile_returns_spread_time_series seems failed with dataframe input

HereticSK:
let `a` to be the output of the recursive call, and the replace the corresponding element in list `ax` with it, so that `a.get_ylim()` have the right method :
```python
        for (i, a), (name, fr_column) in zip(enumerate(ax), mean_returns_spread.iteritems()):
            stdn = None if std_err is None else std_err[name]
            a = plot_mean_quantile_returns_spread_time_series(fr_column,
                                                                   std_err=stdn,
                                                                   ax=a)
            ax[i] = a
            curr_ymin, curr_ymax = a.get_ylim()
```

### DIFF
--- a/alphalens/plotting.py
+++ b/alphalens/plotting.py
@@ -487,11 +487,12 @@ def plot_mean_quantile_returns_spread_time_series(mean_returns_spread,
             ax = [None for a in mean_returns_spread.columns]
 
         ymin, ymax = (None, None)
-        for a, (name, fr_column) in zip(ax, mean_returns_spread.iteritems()):
+        for (i, a), (name, fr_column) in zip(enumerate(ax), mean_returns_spread.iteritems()):
             stdn = None if std_err is None else std_err[name]
-            plot_mean_quantile_returns_spread_time_series(fr_column,
-                                                          std_err=stdn,
-                                                          ax=a)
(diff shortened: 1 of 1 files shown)


twiecki:
@luca-s Can you take a look at this? No worries if you don't have time of course.

alphalens/plotting.py
```diff
@@ -487,11 +487,12 @@ def plot_mean_quantile_returns_spread_time_series(mean_returns_spread,
...
         ymin, ymax = (None, None)
-        for a, (name, fr_column) in zip(ax, mean_returns_spread.iteritems()):
+        for (i, a), (name, fr_column) in zip(enumerate(ax), mean_returns_spread.iteritems()):
             stdn = None if std_err is None else std_err[name]
-            plot_mean_quantile_returns_spread_time_series(fr_column,
-                                                          std_err=stdn,
-                                                          ax=a)
+            a = plot_mean_quantile_returns_spread_time_series(fr_column,
+                                                                   std_err=stdn,
+                                                                   ax=a)
```
luca-s: indentation needs fix

luca-s:
@HereticSK if you could fix indentation at lines 493 and 494 then it's fine. Thank you!

PR #118, opened 2017-01-01
BUG: demeaning in average_cumulative_return_by_quantile

luca-s:
demeaning option returns wrong values if the factor asset universe
varies from date to date. The tests don't cover this scenario, so
this patch fix the issue and add a test case too.

### DIFF
--- a/alphalens/tests/test_performance.py
+++ b/alphalens/tests/test_performance.py
@@ -406,25 +406,8 @@ class PerformanceTestCase(TestCase):
         dr.name = 'date'
         tickers = ['A', 'B', 'C', 'D']
         r1, r2, r3, r4 = (1.25, 1.50, 1.00, 0.50)
-        prices = DataFrame(index=dr, columns=tickers,
-                           data=[[r1**1,  r2**1,  r3**1,  r4**1 ],
-                                 [r1**2,  r2**2,  r3**2,  r4**2 ],
-                                 [r1**3,  r2**3,  r3**3,  r4**3 ],
-                                 [r1**4,  r2**4,  r3**4,  r4**4 ],
-                                 [r1**5,  r2**5,  r3**5,  r4**5 ],
-                                 [r1**6,  r2**6,  r3**6,  r4**6 ],
-                                 [r1**7,  r2**7,  r3**7,  r4**7 ],
-                                 [r1**8,  r2**8,  r3**8,  r4**8 ],
-                                 [r1**9,  r2**9,  r3**9,  r4**9 ],
-                                 [r1**10, r2**10, r3**10, r4**10],
-                                 [r1**11, r2**11, r3**11, r4**11],
-                                 [r1**12, r2**12, r3**12, r4**12],
-                                 [r1**13, r2**13, r3**13, r4**13],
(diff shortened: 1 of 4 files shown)


alphalens/tests/test_performance.py
```diff
@@ -444,3 +444,69 @@ def test_average_cumulative_return_by_quantile(self, before, after,
...
+                             [ 0.0,  0.0,  0.0,  0.0],
+                             [ 0.0,  0.375,  0.90625,   1.664062],
+                             [ 0.0,  0.0,  0.0,  0.0]]),
+                           (0, 3, True, 2,
+                            [[ 0.0, -0.3125, -0.640625, -1.050781],
+                             [ 0.0,  0.0,  0.0,  0.0],
+                             [ 0.0,  0.3125,  0.640625,  1.050781],
+                             [ 0.0,  0.0,  0.0,  0.0]]),
+                          ])
+    def test_average_cumulative_return_by_quantile_2(self, before, after,
```
twiecki: I think this could use a doc-string to describe what it is testing exactly.

alphalens/tests/test_performance.py
```diff
@@ -444,3 +444,69 @@ def test_average_cumulative_return_by_quantile(self, before, after,
...
+                                                   demeaned, quantiles,
+                                                   expected_vals):
+        dr = date_range(start='2015-1-15', end='2015-1-25')
+        dr.name = 'date'
+        tickers = ['A', 'B', 'C', 'D', 'E', 'F']
+        r1, r2, r3, r4 = (1.25, 1.50, 1.00, 0.50)
+        prices = DataFrame(index=dr, columns=tickers,
```
twiecki: Should maybe have a simple function to create `data`: `[[r1**i, r2**i, ...] for i in range(100)]]`

alphalens/tests/test_performance.py
```diff
@@ -444,3 +444,69 @@ def test_average_cumulative_return_by_quantile(self, before, after,
...
+
+        factor, forward_returns = get_clean_factor_and_forward_returns(factor,
+                                                                       prices,
+                                                                       periods=range(0,after+1),
+                                                                       filter_zscore=False)
+        quantile_factor = quantize_factor(factor, by_group=False, quantiles=quantiles)
+        avgrt = average_cumulative_return_by_quantile(quantile_factor, prices, before, after, demeaned)
+        arrays = []
+        for q in range(1, quantiles+1):
+            arrays.append( (q,'mean') )
```
twiecki: non-pep8 formatting: `arrays.append(q, 'mean'))`

alphalens/tests/test_performance.py
luca-s: good idea

alphalens/tests/test_performance.py
luca-s: I thought about that but I didn't want to make the code too obscure. But I can change that if you like.

twiecki: Yeah, I'm not too sure myself. I would prefer the list comprehension because 1) less code duplication, 2) less wall-of-text.

luca-s:
it should be fine now

twiecki:
:+1:

PR #119, opened 2017-01-01
Rearranged Group Specific Breakdown plots

luca-s:
Mean Return and Average Cumulative Returns plots are now side by side to make the most of them, as the former is a detailed view of the latter at specific periods. Average Cumulative Returns plot is still optional.

untitled


twiecki:
Nice -- looks way cleaner!

Issue #80, opened 2016-08-31
Create fake factor that's very predictive


twiecki:
For demonstrating the capabilities and what it would like if a factor was predictive, we should include made-up data of such a factor with look-ahead bias.

luca-s:
It should be straightforward to modify alphalens/examples/tear_sheet_walk_through.ipynb so that the factor looks at future prices instead of past ones (one line of code change). Alternatively, you could keep the current factor but shift it back in time so that everyday you know the factor of tomorrow. Currently the factor is shifted forward one day to avoid Look-Ahead bias, but we want to introduce the bias to have a predictive factor, so we do the opposite (still one line of code change).

twiecki:
yep, that would be a good way to do it. want to do a PR?

luca-s:
No really, I was just suggesting that because I use the trick to do my own testing sometime.

twiecki:
Fair enough, thanks for the suggestion!

jameschristopher:
fixed https://github.com/quantopian/alphalens/pull/121

PR #121, opened 2017-01-04
New example NB: predictive vs. non-predictive factor

luca-s:
This is a very simply notebook in which we use Alphalens to analyse a predictive and a non-predictive factor. The goal is to have exemplifying outputs for both scenarios to make clear what to expect from Alphalens in the case of a predictive and a non-preditive factor

This is loosely related to #80


twiecki:
This looks awesome!

Issue #124, opened 2017-01-09
Yellow colour is hard to see on cumulative returns by quantile plots.


jmccorriston:
![image](https://cloud.githubusercontent.com/assets/4500507/21773567/99904e02-d65d-11e6-8ef2-06e823265c0c.png)

I didn't even see the lines for Quantile 2!

DMTSource:
I came to post the same issue but I just noticed on the front page for this repo there is a nice gray bg color for each subplot. These are not shown here(or in my Research notebook on Quantopian) when generating a tearsheet. The code in plotting.py seems to be setting the colors via the rc defaults(line 85).

Perhaps the yellow can stay to retain color variety for higher number of quantiles(assuming this was part of the intent cause there are only so many differentiable colors in mpl) and the bg color can be re-addressed?

luca-s:
I agree with @DMTSource that making the background dark would solve the problem. 

By the way, I believe we should use seaborn.axes_style() to change the background colour, while the current code uses seaborn.plotting_context() that is intended to control the **scale** of plot elements.

luca-s:
Fixed in #125, I am closing this. Feel free to re-open it if required

PR #125, opened 2017-01-14
Fix issue 124

luca-s:
Switched to 'darkgrid' axes style to improve readability, issue #124 

Also this fix another issue. The plotting.context function tries to set 'axes.facecolor' and 'figure.facecolor' properties using seaborn.plotting_context(), but that cannot work. seaborn.plotting_context() is intended to control the scale of plot elements while seaborn.axes_style() fucntion should be used for the style. So a combination of the two is required to have full control of the plotting properties. See http://seaborn.pydata.org/tutorial/aesthetics.html

### DIFF
--- a/alphalens/plotting.py
+++ b/alphalens/plotting.py
@@ -30,14 +30,13 @@ from functools import wraps
 DECIMAL_TO_BPS = 10000
 
 
-def plotting_context(func):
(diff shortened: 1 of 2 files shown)


twiecki:
LGTM, thanks!

PR #128, opened 2017-01-19
Improved performance of utils.common_start_returns

luca-s:
I ran some benchmarks and it was too slow.

### DIFF
--- a/alphalens/utils.py
+++ b/alphalens/utils.py
@@ -275,7 +275,10 @@ def common_start_returns(factor, prices, before, after,
         -before to after
     """
 
-    returns = prices.pct_change(axis=0)
+    if cumulative:
+        returns = prices
+    else:
+        returns = prices.pct_change(axis=0)
 
     all_returns = []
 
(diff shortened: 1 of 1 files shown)


luca-s:
I'll go on and merge this, as this is a pretty small change and well covered by the tests

alphalens/utils.py
```diff
@@ -292,28 +295,31 @@ def common_start_returns(factor, prices, before, after,
         ending_index = min(day_zero_index + after + 1, 
                            len(returns.index))
 
-        series = returns.iloc[starting_index:ending_index, :]
+        equities_slice = set(equities)
+        if demean is not None:
+            demean_equities = demean.loc[timestamp].index.get_level_values('asset')
+            equities_slice |= set(demean_equities)
+
+        series = returns.ix[starting_index:ending_index, equities_slice]
```
twiecki: Should use `.loc`, `.ix` is deprecated.

luca-s: Oh yes, pandas 0.20.0 deprecates it. I usually prefer .loc and .iloc but this time I really needed to use .ix because it  supports mixed integer and label based access.The alternate version with an .iloc followed by a .loc is slower. I will change the code anyway. sorry about that.

PR #130, opened 2017-01-20
WIP - new api

jameschristopher:
Here is where I'm headed with the new api thoughts from https://github.com/quantopian/alphalens/pull/110 so to review creating a tear sheet would be a two step process.

1. format the data
2. call for the tear sheet 

where all of the tear sheets take a `factor_data` dataframe that looks something like 
screen shot 2017-01-13 at 11 53 54 am

Right now the main blocking issue is the event-study-esque plots as that [function](https://github.com/quantopian/alphalens/blob/master/alphalens/performance.py#L469) requires raw prices. I think that the number of plots and the uniqueness of what they are saying probably merits them getting a separate tear sheet (which would be able to take raw prices).

### DIFF
--- a/alphalens/performance.py
+++ b/alphalens/performance.py
@@ -349,7 +349,7 @@ def factor_rank_autocorrelation(factor_data, period=1):
     to account for systematic shifts in the factor values of all names or names
     within a group. This metric is useful for measuring the turnover of a
     factor. If the value of a factor for each name changes randomly from period
-     to period, we'd expect an autocorrelation of 0.
+    to period, we'd expect an autocorrelation of 0.
 
     Parameters
     ----------
--- a/alphalens/performance.py
+++ b/alphalens/performance.py
@@ -446,17 +446,3 @@ def average_cumulative_return_by_quantile(quantized_factor, prices,
         return pd.DataFrame( {'mean': q_returns.mean(axis=1), 'std': q_returns.std(axis=1)} ).T
 
     return quantized_factor.groupby(quantized_factor).apply(average_cumulative_return)
-
-
-def compute_information_statistics(factor_data,
-                                   group_adjust=False,
-                                   by_group=False):
-
-
-    ic = factor_information_coefficient(factor_data, group_adjust, by_group)
-
-    mean_monthly_ic = mean_information_coefficient(factor_data,
-                                                   group_adjust,
-                                                   by_group,
-                                                   "M")
-    return ic, mean_monthly_ic
--- a/alphalens/performance.py
+++ b/alphalens/performance.py
@@ -23,8 +23,7 @@ from statsmodels.tools.tools import add_constant
 from . import utils
 
 
-def factor_information_coefficient(factor,
-                                   forward_returns,
+def factor_information_coefficient(factor_data,
                                    group_adjust=False,
                                    by_group=False):
     """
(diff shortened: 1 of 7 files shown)


twiecki:
This looks like a great start. Did you start from recent master? Seems like there are conflicts.

luca-s:
Your changes look good to me. I agree with you that the event-study-esque plots could go in a separate tear sheet. Still dropping `prices` after the initial data formatting process means we are losing information and it might become a problem for the development of future features. But this is just a minor issue and It might not be a problem at all, I just wanted to highlight this. 

A side note regarding the data, we are probably not following the tidy data guidelines you mentioned in #110 (thanks for the article, that was interesting). But we can fix it later on I guess, if we really want to do it.

twiecki:
@luca-s Which part of tidy data are we missing? The fact that the forward returns periods should actually be a hierarchical index?

luca-s:
My understanding is that the forward returns periods should not be columns headers but rather values in a  dedicated column. So we would have only two columns for the forward returns: a column whose values are the number of periods (1, 5, 10 or whatever) and another column with the forward return values for each periods.

twiecki:
yes, except that your first column should also be an index, I think. Btw. now is the time to think about this.

jameschristopher:
@twiecki I opened the branch last Friday so I think I'm about 5 commits behind. 

Luca I'm not so sure, the author defines a value as 

> Value: The actual measurement or attribute. 152 cm, 80 kg, female, etc.

The number of periods are not a measured attribute of the factor, they are part of the measurement process of forward returns. So I'd argue that they should be incorporated into the column header.

luca-s:
@jameschristopher it seems to me the example of the 'Income" column really matches our use case. Also I believe the number of periods should be a column because I have already faced the problem of needing the number of periods as a parameter and I ended up reading it from the column name. That cannot be tidy data :) If you think about it, we don't have that information in our DataFrame expect for the column names. Maybe those are weak points but I just wanted to make them clear anyway so we can take the best decision.

jameschristopher:
Those are weak points. That is an excellent point and something I've also thought about. A common issue that I ran into with this PR is getting the forward returns columns over and over again. You can see in the code that I use a regex to do this - not so elegant.

twiecki:
I think it should definitely be an index-level (or column, but it's not really data, it's an identifier). Not separate columns as it is now.

jameschristopher:
Thomas can you provide an illustration of what you mean? Something like this?
screen shot 2017-01-24 at 7 20 08 am

twiecki:
@jameschristopher exactly like that.

jameschristopher:
yes as would 'group' and 'quantile'

jameschristopher:
After some more thought I'd like to submit a multilevel column as the approach we should take
screen shot 2017-02-02 at 4 11 43 pm

twiecki:
I think that's the best solution so far. @luca-s do you agree?

luca-s:
That's fine with me. We have already had the chance to discuss all the relevant points, so I certainly trust James choice if he thinks that's the best way to go. Let's go with that then.

twiecki:
Did you remove the hierarchical column again? screen shot 2017-02-13 at 08 58 25

twiecki:
But I think otherwise this looks great. @luca-s can you take a quick look too?

luca-s:
Sure, I will.

luca-s:
Great work, I like the new api. 

I only have few comments:

1 - 'quantile_factor.max()'  (used to fetch the number of quantiles/bins) should be replaced with 'int(quantile_factor.max())' because quantiles/bins values might be float, depending on quantiles/bins arguments and that would break the code.

2 - We don't have quantile statistics anymore. I found them very useful when using non default quantiles/bins arguments.

screenshot from 2017-02-13 12-15-45

3 - Could you please add the call to 'create_event_returns_tear_sheet' in the 'predictive_vs_non-predictive_factor' NB? As the purpose of that NB is to show what to expect in each plot in the cases of  predictive and non-predicvive fators it would be useful to have all the plots there.

4- I miss to have mean return and event style return plots side by side, but there is no much we can do about it.

screenshot from 2017-02-13 12-45-04

luca-s:
I noticed another little issue at [here](https://github.com/quantopian/alphalens/blob/new-api/alphalens/utils.py#L306). When calling `util.quantize_factor` it is possible that some of the factor values are dropped, depending on the quantiles/bins arguments. This is a feature, as it could be useful to drop top/bottom values as a measure to remove outliers. So we don't want to keep factor entries whose values are NaN, this means a second call to 
`merged_data = merged_data.dropna()` is required after calling `util.quantize_factor`

jameschristopher:
Yes Thomas the hierarchical columns weren't workable.

If there are no further comments I'd like to merge this PR, wait a week to see what bugs come up, fix those bugs and push an official release `v0.1` to PyPi.

twiecki:
This was not an easy one -- thanks @jameschristopher and @luca-s!

luca-s:
thanks @jameschristopher

PR #148, opened 2017-03-07
BUG: create_event_returns_tear_sheet crashes

luca-s:
When custom ranges are provided to quantiles/bins arguments then factor_data['factor_quantile'] contains float instead of integers, that in turn led to crashes


twiecki:
Good catch!

PR #151, opened 2017-03-10
BUG: create_full_tear_sheet doesn't allow to set custom plotting_context/axes_style

luca-s:
For example the following doesn't work:

```
with alphalens.plotting.plotting_context(style='whitegrid'), alphalens.plotting.axes_style():
      alphalens.tears.create_full_tear_sheet(...,  set_context=False)
```

The problem is that each tear sheet function have a decorator, so we need to disable it when calling a tear sheet function from another one.


twiecki:
Yep, that makes sense. Thanks!

Issue #154, opened 2017-05-17
Inconsistent "Returns Analysis" Table


MichaelJMath:
This issue is regarding the "Returns Analysis" table on the tearsheet. More specifically it relates to the last 3 rows in the table:

1. Mean Period Wise Return Top Quantile (bps)
2. Mean Period Wise Return Bottom Quantile (bps)
3. Mean Period Wise Spread (bps)

The issue is in how the return statistics are quoted. The Mean Period Wise Returns, Top Quantile and Bottom Quantile (1 and 2 above) are quoted as a holding period excess return (i.e. ending_value/beginning_value - 1). However, the Mean Period Wise Spread (bps) (3 above) is quoted on a one-period basis (i.e. it uses the **utils.rate_of_return** function to convert holding period return to a one period rate of return)

My recommendation is the following:

1. Make the all return rows quoted using the same method. (Either all one-period returns or all holding period returns). Based on the description, it sounds like they should be quoted as one-period returns.
2. I think it should be made clear to the user that returns are "excess returns", as in the code, it appears the universe mean return is subtracted from the return to come up with the quoted returns. Possibly add this in the description in the first column of the pandas table.

My notes are below for more clarity/brevity:

![al_returns_analysis](https://cloud.githubusercontent.com/assets/28760724/26163341/60481888-3af7-11e7-953a-25479fb9a994.png)

luca-s:
Thank you for noticing and reporting this. Regarding point 1 they were all suppose to be  one-period returns. Regarding point 2, the demeaing is controlled by the long_short flag passed to the different tear sheet functions. The description could be added to the flag documentation.

Would you like to submit a pull request?

MichaelJMath:
I can give it a shot once I learn how. I'm still in the process of learning to use Git and Git Hub.

luca-s:
If you like to provide a PR, then it will be quite easy:

- fork Alphalens project on your github account
- clone your forked Alphalens on your workstation
- create a branch where you'll commit the changes
- push the branch on your forked Alphalens
- from github web page you will see your branch and a "Create pull request" option
- Once you create the PR we'll see that in main Alphalens project and we will be ready to merge the patch or to ask you to provide more commits on that branch

PR #158, opened 2017-06-17
Daily compound returns analysis

MichaelJMath:
In response to Issue #154, I modified the Returns Analysis table so that the "Mean Period Wise Return Top (Bottom) Quantile" rows displayed daily (one-period) returns, so that it is consistent with the "Mean Period Wise Spread" row. 

image


luca-s:
Thank you @MichaelJMath

Issue #156, opened 2017-05-17
perf.factor_alpha_beta doc string typo


MichaelJMath:
I believe there is a mistake in the docstring for **perf.factor_alpha_beta**.

Current doc string:
    
Computes the alpha (excess returns), alpha t-stat (alpha significance),
    and beta (market exposure) of a factor. A regression is run with
    the period wise factor universe mean return as the dependent variable
    and mean period wise return from a dollar-neutral portfolio weighted
    by factor values as the independent variable.

My recommendation (switch dependent and independent variable):
    
Computes the alpha (excess returns), alpha t-stat (alpha significance),
    and beta (market exposure) of a factor. A regression is run with
    the period wise factor universe mean return as the independent variable
    and mean period wise return from a dollar-neutral portfolio weighted
    by factor values as the dependent variable.

luca-s:
I agree with you. Are you interested in providing a PR?

PR #159, opened 2017-06-17
Correct factor_alpha_beta() doc string

MichaelJMath:
This commit correctly identifies the independent and dependent
variable for the regression.

Issue #156


luca-s:
Thank you @MichaelJMath

Issue #155, opened 2017-05-17
Standard Error Period Conversion in tears.create_returns_tear_sheet


MichaelJMath:
In **tears.create_returns_tear_sheet**: it appears the following line of code attempts to convert the holding period standard error to a one-period standard error. 

```python
compstd_quant_daily = std_quant_daily.apply(utils.rate_of_return, axis=0)
```

However, while correct for return period conversions, I believe this conversion method is mathematically incorrect for standard deviation/standard error. The correct approximation is below:

se_h = holding period standard error
se_d = one-period standard error
period = holding period length

se_d = se_h / sqrt(period)

```python
import numpy as np
period_len = std_quant_daily.name
compstd_quant_daily = std_quant_daily.apply(lambda x, n: x / np.sqrt(n), args=(period_len,))
```

To get an exact calculation, one would need the expected value and standard deviation of one-period returns. This would be a bit lengthier to fix. For a discussion on the exact calculation, see https://corporate.morningstar.com/US/documents/MethodologyDocuments/MethodologyPapers/SquareRootofTwelve.pdf

luca-s:
Thank you for looking so carefully at the code. That's obviously a bug and thanks for finding it out. would you be interested in providing a PR?

luca-s:
PR #160 fixed this issue using an approximation. Optionally we can update the conversion function with the precise equation. Not sure we really care about that,

PR #160, opened 2017-06-19
Fix std se conversion

MichaelJMath:
Correct the formula/function used to convert the standard error in the tear sheet to a one-period standard error. This is done by creating a new function in the utils module.

Issue #155


luca-s:
Thank you @MichaelJMath

Issue #152, opened 2017-05-03
Better error message for not enough quantiles for factor values


CaptainKanuk:
Factors which draw from small discrete sets (such as a buy/sell factor with -1, 0, 1) fail in alphalens. A fairly uninformative error is generated. For repro you can try running it on a factor constructed from this sentiment data. We should have a more elegant failure or some kwarg to tell alphalens the factor is discrete. I solved the issue locally by adding in a tiny amount of random noise to my factor, effectively making it continuous.

https://www.quantopian.com/data/sentdex/sentiment

luca-s:
For factors containing discrete values you can use the 'bins' argument that force alphalens to perform a binning on the values instead of creating quantiles. With discrete values from your example (-1,0,1) you could try something like:

```
bins=[-1.5, -0.5, 0.5, 1.5] 
quantiles=None 
```
For details see [here](https://github.com/quantopian/alphalens/issues/105)

Regarding a more informative error,I agree with you.

CaptainKanuk:
Got it, thanks. This is helpful and having a more useful error, or a kwarg for discrete data, would be awesome.

mmargenot:
I have run into this same issue before with continuous factors that simply have a lot of 0 values. I have solved this similarly to Delaney, by adding a very small amount of random noise. In this case, I don't think specifying the bins will solve the issue since I do not know the quantiles.

luca-s:
@mmargenot  I would consider your issue different from the one reported by @CaptainKanuk. The problem you faced is due to pd.qcut (the function that calculate the quantiles) not being able to provide a workaround when identical values span more than one quantile. Interesting to note, in pandas 0.20.0 they added an option (duplicates='raise'|'drop') that fixes that, which would result in less bins than specified, and some larger bins (with more elements) than others. We could make use of that option eventually.

Until then, another workaround could be to use bins=5 and quantiles=None, that chooses the bins/quantiles to be evenly spaced according to the values themselves, while quantiles=5 chooses the bins/quantiles so that you have the same number of records in each bin/quantile.

luca-s:
Closing this

PR #161, opened 2017-06-20
Better error message for not enough quantiles for factor values

luca-s:
Fix issue #152


twiecki:
Why the use of a decorator if it's just in a single function?

luca-s:
The other option would have been to wrap the whole function body around a try/catch, plus the endless message. I thought that would have made the function hard to read and ugly.

twiecki:
I see, I think this looks good!

twiecki:
We should cut a new release.

PR #164, opened 2017-07-03
Added even study Notebook example

luca-s:
Even though Alphalens purpose is not to be an even study like tool, it is still quite straightforward to use its functions to achieve the purpose. Also, because some users asked for this kind of usage on Quantopian I thought to add this example usage.


twiecki:
Asked @ttrivedi to review.

twiecki:
Looks great, thanks @luca-s!

twiecki:
Still need to add it to `docs/sources/examples.rst`.

luca-s:
@twiecki what is docs/sources/examples.rst file? I cannot see it

twiecki:
Sorry, I meant https://github.com/quantopian/alphalens/blob/master/docs/source/index.rst.

luca-s:
Actually I didn't create new API so there is nothing to add the file. Am I wrong?

I simply used existing Alphalens functions to create an event study, but we can still add a new API if you like, e.g. alphalens.tears.create_event_study_tear_sheet

luca-s:
I added the function  alphalens.tears.create_event_study_tear_sheet, as a prototype. if you like it I can refine the code and fix the documentation. Otherwise I can simply revert the commit

twiecki:
@luca-s looks great, I would definitely adapt the NB to use the new functionality.

luca-s:
Good, then I will refine the code, move some functionalities in separate functions, add documentation and update the NB.

luca-s:
Done, the code is now nicer and the NB has been updated.

twiecki:
This is excellent. The NB still needs to be added to https://github.com/quantopian/alphalens/blob/master/docs/source/index.rst and some basic tests would be great.

luca-s:
Uhmm, I cannot see where I should add the NB. The other example NBs are not there. 

Regarding the tests the new functions uses existing functionalities that already have test cases.

twiecki:
Oh, sorry, we set this up for every project but this. This is all set. Thanks!

PR #168, opened 2017-07-17
Remove old tearsheet and deprecation warnings

jameschristopher:
It has been 5 months and 2 releases since the API rework, judging by activity on Quantopian's forums all the notebooks I viewed on a quick cursory glance seem to be using the new API, in addition by the time Alphalens on Quantopian gets updated even fewer users will be using the old API. I think the next release would be an appropriate time to remove the deprecated code.


twiecki:
Awesome, thanks @jameschristopher! Hope all is well.

Issue #94, opened 2016-10-06
Create breaking test on tear-sheet


twiecki:
Add a simple test that runs the tear-sheet and makes sure it does not raise an exception.

PR #169, opened 2017-07-18
Issue #94

luca-s:
Add a simple test that runs the tear-sheet and makes sure it does not raise an exception


twiecki:
Super useful, thanks.

Issue #172, opened 2017-07-24
"Create Tearsheet" Functions - Lack of Clarity: long_short vs. demeaned 


MichaelJMath:
In working on issue #167, I noticed a lack of clarity in regards to the use of the `long_short` vs. `demeaned` flags in some of the functions. The input parameters of the "create tearsheet" functions in the tears module contains the `long_short` parameter. This bool parameter is then used within those functions as the demean input to functions like `perf.mean_return_by_quantile()`. 

In my understanding, the `demeaned` argument is used to basically subtract each return by the mean universe return. This is different than when it is used in the call to `perf.factor_returns()` which factor weights the returns. If `long_short` is True, then the factor "values" themselves, not the returns, are demeaned during the weighting process.

I have thought of two possible ways of dealing with this:

1. We could modify the doc string so that the `long_short` parameter description more accurately describes this issue. Something like: 
``` python
"""
long_short : bool
        Should this computation happen on a long short portfolio? If so, then factor values
        will be demeaned across the factor universe when factor weighting the portfolio. 
        Additionally, mean quantile returns will be demeaned across the factor universe.
"""
```
2. We could create a separate `demeaned` argument for these "create tearsheet" functions.

luca-s:
Sounds good to me, I would go with option 1

PR #174, opened 2017-07-25
Long short docstring

MichaelJMath:
Clarified long-short docstring in tears module for create tearsheet functions (issue #172 ).


luca-s:
Many thanks @MichaelJMath

PR #183, opened 2017-08-09
Returns Decomposition

eigenfoo:
Closes #180 

A rebase will be necessary for this PR, due to #179

PR #185, opened 2017-08-25
ENH returns decomposition

eigenfoo:
Identical to #183. Opening a new PR both to avoid a rebase, and also to make things tidier and easier for everyone.


eigenfoo:
conda can't seem to find empyrical. Not sure what this problem really means, or how to fix it. @twiecki?

```
PackageNotFoundError: Packages missing in current channels:
            
  - empyrical

We have searched for the packages in the following channels:
            
  - https://repo.continuum.io/pkgs/free/linux-64
  - https://repo.continuum.io/pkgs/free/noarch
  - https://repo.continuum.io/pkgs/r/linux-64
  - https://repo.continuum.io/pkgs/r/noarch
  - https://repo.continuum.io/pkgs/pro/linux-64
  - https://repo.continuum.io/pkgs/pro/noarch
            
The command "conda create -q -n testenv --yes python=$TRAVIS_PYTHON_VERSION ipython flake8 numpy scipy nose matplotlib pandas statsmodels seaborn empyrical collections" failed and exited with 1 during .
Your build has been stopped.
```

eigenfoo:
I think I fixed the problem, but now there's another bug

```
AttributeError: 'dict' object has no attribute '__NUMPY_SETUP__'
The command "python setup.py install" failed and exited with 1 during .
```

I really have no idea what this means

.travis.yml
```diff
@@ -24,7 +24,7 @@ before_install:
   - cp alphalens/tests/matplotlibrc .
 
 install:
-  - conda create -q -n testenv --yes python=$TRAVIS_PYTHON_VERSION ipython flake8 numpy scipy nose matplotlib pandas statsmodels seaborn
+  - conda create -q -n testenv --yes python=$TRAVIS_PYTHON_VERSION ipython flake8 numpy scipy nose matplotlib pandas statsmodels seaborn
```
twiecki: stray space

twiecki:
I restarted the build, hopefully it's intermittent.

twiecki:
The failure is still present so nothing intermittent.

luca-s:
Hi @eigenfoo,  What is the status of this PR? What is still missing?

twiecki:
@luca-s Probably the first would be figuring out the NUMPY bug.

luca-s:
That seems a Bottleneck-1.2.1 setup issue

luca-s:
That seems a Bottleneck-1.2.1 setup issue

Bottleneck is required by empyrical and empyrical is used by pyfolio too, but pyfolio works. So a comparison between pyfolio setup with alphalens setup should help

twiecki:
@luca-s Good catch that seems likely.

eigenfoo:
@luca-s @twiecki sorry for the late response, I just saw this while clearing my notifications.

The PR was almost ready for merge, and just needed a green light from engineering. That being said though, there are still improvements to be made:

- Add tests for `calc_vifs` and `decompose_returns`: as these functions actually perform analyses and compute numbers, it's important that they're well-tested. Preferably, we'd write the tests before merging this PR, but that's up to the core developers.
- Add documentation (maybe an example notebook?) to alphalens to explain to the user base what the multifactor alpha is, why its useful, and how to interpret it. I believe that @cfenaroli worked on a [case study somewhere here](https://github.com/quantopian/research_public) that would be perfect, but I can't recall where.

mmargenot:
I have his most recent draft of the case study floating around and can provide it easily if you need the example.

mmargenot:
Running `python setup.py install` for both `empyrical` and this branch of `alphalens` yield that same bug with `__NUMPY_SETUP__`. Flagging down this line https://github.com/kwgoodman/bottleneck/blob/master/setup.py#L24 in Bottleneck-1.2.1.

eigenfoo:
@mmargenot done!

mmargenot:
Hm. We seem to have broken the connection to travis. I'll poke around.

mmargenot:
@eigenfoo We have connection again. Travis is still a little cranky, though.
```
alphalens/performance.py:481:1: W293 blank line contains whitespace
alphalens/performance.py:638:1: W391 blank line at end of file
alphalens/plotting.py:968:80: E501 line too long (83 > 79 characters)
```

eigenfoo:
@mmargenot thankfully they're just pep8 fixes!

alphalens/performance.py
```diff
@@ -527,3 +529,109 @@ def average_cumulative_return(q_fact, demean_by):
...
+        https://en.wikipedia.org/wiki/Variance_inflation_factor
+    '''
+    if hierarchy is None:
+        hierarchy = risk_factors.columns.values
+
+    # construct a cumulative hierarchy
+    cum_tiers = []
+    for i in range(len(hierarchy.values())):
+        cum_tiers.append([item for a in range(0, i+1)
+                          for item in hierarchy.values()[a]])
```
mmargenot: Getting dictionary keys and values changed in Python 3. I think `hierarchy.values()` in line 603 should be replaced with `list(hierarchy.values())` when you try to access it with brackets.

alphalens/performance.py
```diff
@@ -527,3 +529,109 @@ def average_cumulative_return(q_fact, demean_by):
...
+    if hierarchy is None:
+        hierarchy = risk_factors.columns.values
+
+    # construct a cumulative hierarchy
+    cum_tiers = []
+    for i in range(len(hierarchy.values())):
+        cum_tiers.append([item for a in range(0, i+1)
+                          for item in list(hierarchy.values())[a]])
+    cum_hierarchy = OrderedDict()
+    cum_hierarchy.update((hierarchy.keys()[i], cum_tiers[i])
```
mmargenot: `hierarchy.keys()` -> `list(hierarchy.keys())`

alphalens/performance.py
```diff
@@ -527,3 +529,109 @@ def average_cumulative_return(q_fact, demean_by):
...
+    idx = algo_returns.index.intersection(risk_factors.index)
+    risk_factors = risk_factors.loc[idx]
+    algo_returns = algo_returns.loc[idx]
+
+    idx = np.append(['Alpha'], risk_factors.columns.values)
+    betas = pd.DataFrame(index=idx, columns=hierarchy.keys())
+    returns_decomposition = pd.DataFrame(index=idx, columns=hierarchy.keys())
+    t_stats = pd.DataFrame(index=idx, columns=hierarchy.keys())
+
+    for tier, factors in cum_hierarchy.iteritems():
```
mmargenot: I think we can get away with changing `.iteritems()` to `.items()`. This operation seems to be bounded by the number of risk factors, which should usually be a reasonable number.

mmargenot:
I just got all tests with this running on my local machine. There might be something weird with how Travis is set up.

twiecki:
@luca-s If you have time, want to take a look at this?

luca-s:
Sure, I'll have a look. The branch needs to be rebased on master though, there are conflicts.

alphalens/tears.py
```diff
@@ -393,6 +395,22 @@ def create_returns_tear_sheet(factor_data,
                                            ylim_percentiles=(5, 95),
                                            ax=ax_quantile_returns_bar_by_group)
 
+    ff_factors = ep.utils.load_portfolio_risk_factors() \
+        .drop(['RF'], axis='columns')
```
luca-s: Can  ep.utils.load_portfolio_risk_factors throw? What happen if you run Alphalens without network connection? I would embed the code in a try catch statement and let Alphalens compute the normal tearsheet if the risk factor data is not available

alphalens/tears.py
```diff
@@ -393,6 +395,22 @@ def create_returns_tear_sheet(factor_data,
...
+    hierarchy = OrderedDict([
+        # not typos! These are the Fama-French factors names from Dartmouth
+        ('Market', ['Mkt-RF']),
+        ('Style', ['SMB', 'HML', 'Mom'])
+    ])
+
+    factor_returns.index = factor_returns.index.tz_localize('UTC')
```
luca-s: Alphalens code can deal with both tz_localized and non-tz_localized input. This statement force factor_returns to be tz naive because it breaks when factor_returns is tz_localized already.

luca-s:
This branch is too far behind, I'll rebase on master

mmargenot:
@luca-s @twiecki would it make sense include a flag to ask for a returns decomposition or not? The risk factors come in from `empyrical` tz-aware and it seems reasonable to me that in the case where someone wants a returns decomposition they would be using tz-aware inputs.

luca-s:
What about checking that factor has the same time zone information as risk factor ( `factor.index.levels[0].tz == ff_factors.index.tz`) ? If they differ then we can print a warning message detailing the issue and skip the returns decomposition. This way the user is less annoyed because alphalens can continue the analysis and it is also aware of the problem and can fix it if they like to.

alphalens/tears.py
```diff
@@ -393,6 +395,24 @@ def create_returns_tear_sheet(factor_data,
...
+        ('Market', ['Mkt-RF']),
+        ('Style', ['SMB', 'HML', 'Mom'])
+    ])
+    if factor_returns.index.tzinfo is None:
+        factor_returns.index = factor_returns.index.tz_localize('UTC')
+    else:
+        factor_returns.tz_convert('UTC')
+    for i, period in enumerate(factor_returns.columns.values):
+        returns_decomposition = \
+            perf.decompose_returns(factor_returns.iloc[:, i],
```
luca-s: I have just realized that passing 'factor_returns' to `perf.decompose_returns` works only in the very specific scenario (thought very common and probably the most common scenario), that is when the factor is:
- computed on a daily basis
- the period is 1 day
- the entry/exit prices for the factor are market open

luca-s: I have just realized that passing factor_returns to perf.decompose_returns works only in a very specific scenario (thought very common and probably the most common scenario), that is when the factor is:
computed on a daily basis
the period is 1 day
the entry/exit prices for the factor are market open

mmargenot: Good catch!

Unfortunately, I don't think it's possible to do the performance attribution on a more granular time scale than 1 day due to the availability of the Fama French without computing the factors ourselves.

luca-s:
I've started having some doubts regarding this new feature recently, I am not so sure anymore it integrates well in the current Alphalens analysis and I'd like to hear what the other people think of that.

Here are my major concerns:
- the return decomposition analysis requires external data that has to be downloaded (the alpha/beta computation currently done on Alphalens doesn't require that). This is mitigated if we let Alphalens skips the returns decomposition when the data is not available, but still it's a shift in Alphalens behavior that shouldn't be underestimated
- currently Alphalens looks only at the predictive power of an alpha factor and this "limitation" in scope makes it very flexible: there are no assumpions on the market (US, Europe, developing, etc. ) or on the trading frequency (intraday, daily, weekly) and probably on other things that I cannot think of at the moment. This new feature changes that but again we can still  mitigate the problem letting Alphalens skip the returns decomposition when the input data doesn't conform to the requirements.
- More generally I believe this type of analysis suits better pyfolio than Alphalens

I wonder if it wouldn't make more sense to use the current code to compute the risk factors returns and exposure and use this data to call `pyfolio.create_perf_attrib_tear_sheet` whose job is to perform exactly this kind of analysis. As Alphalens has recently gained the ability to create input data suitable for pyfolio, I believe the path explained above better suits and better makes use of both Alphalens and Pyfolio features without mixing them and enforcing a boundary between the two tools.

alphalens/tears.py
```diff
@@ -417,6 +419,24 @@ def create_returns_tear_sheet(factor_data,
         plt.show()
         gf.close()
 
+    if factor_returns.index.tzinfo is None:
+        warnings.warn("to do a risk factor performance attribution,"
+                      "you must pass in a tz-aware Series")
+        return
```
luca-s: Instead of returning, I believe the analysis could continue skipping return decomposition

alphalens/tears.py
```diff
@@ -417,6 +419,24 @@ def create_returns_tear_sheet(factor_data,
         plt.show()
         gf.close()
 
+    if factor_returns.index.tzinfo is None:
```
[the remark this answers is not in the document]
mmargenot: If they're different, it's not as big of a deal, no? pandas can handle differing time zones, so the `intersection` that occurs later is still safe. If returns are in `EST` and the factors are in `UTC`, then as long as the datetimes still match up, e.g. 9AM UTC and 4AM EST, it will work. The issue is whether the `factor_returns` are totally naive, which is what we want to check for.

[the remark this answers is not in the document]
mmargenot: Though a check later to make sure that they DO match up, like the market open issue that you mentioned above, would be good.

twiecki:
Closing as we can use pyfolio now for risk analysis.

luca-s:
I wonder if we should eventually incorporate the return decomposition functions inside alphalens.performance  or alphalens.utils. They might come in handy if a user has their own risk factors and they want to compute the factor loadings, which is required by Pyfolio risk analysis. Or  we could incorporate those utility functions inside Pyfolio maybe ?

This is just an idea anyway, I haven't thought thoroughly on how to compute the factor loadings from custom factor risks

Issue #187, opened 2017-09-19
How to interpret the cumulative return for N forward period in cumulative factor return plots?


HereticSK:
In `plot_cumulative_returns_by_quantile`, cumulative returns for periods more than 1 is computed as:
```Python
 if period > 1:
        ret_wide = pd.rolling_apply(
            ret_wide,
            period,
            # compound returns
            lambda ret, period: ((np.nanmean(ret) + 1)**(1. / period)) - 1,
            min_periods=1,
            args=(period,))

    cum_ret = ret_wide.add(1).cumprod()
```
It converts the N-period return to one-period return by taking geometric mean. But I don't know how to make sense of this "derived" return, because:
- Numerically, this "derived" one-period return is not equal to the exact one-period return computed as Price_t/Price_t-1 - 1. Maybe it can be seen as some kind of approximation, but that requires we make some assumption on the N-period return
- Intuitively, cumulative return should be interpreted as generated by some continuously rebalanced portfolio. For instance, cumulative return using simple one-period returns can be interpreted as if the factor weighted portfolio is rebalanced every period. This is very intuitive. But if the cumulative return is computed using the "derived" one-period returns, the interpretation is not as intuitive.

It seems to me that the concept of cumulative return only applies to one-period returns. It does not make much sense to derive some kind of one-period returns from N-period returns just to compute the cumulative return. How much value does it provide for assessing the factor?

luca-s:
The "derived" one-period return is what is called the one-period rate of return:  the returns the factor quantile would have had every 1-period if it had grown at a steady rate. 

But why does Alphalens computes the 1-period rate of return instead of computing the cumulative returns from a N-periods returns? Alphalens makes an assumption on how the factor should be used, that is how and how often a portfolio should be  rebalanced.

If the factor has the best performance at N periods (N >1) then a naive approach of using that factor would be to rebalance the portfolio every N periods. That would work but also it would make the portfolio performance dependent on the specific starting point. Also, Alphalens statistics are calculated considering the factor performance at each period, but if the factor is traded every N periods (rebalancing every N periods) we cannot expect to achieve the same performance.

[...]

PR #188, opened 2017-09-21
Updated functions documentation: related to #187

luca-s:
Explained how to interpret the cumulative return for N forward period in
performace.plot_cumulative_returns and performance.plot_cumulative_returns_by_quantile

### DIFF
--- a/alphalens/plotting.py
+++ b/alphalens/plotting.py
@@ -704,6 +704,12 @@ def plot_cumulative_returns(factor_returns, period=1, ax=None):
     """
     Plots the cumulative returns of the returns series passed in.
 
+    When 'period' N is greater than 1 the cumulative returns plot is computed
+    building and averaging the returns of N portfolios started at subsequnet
+    periods (1,2,3,...,N) and where each portfolio is then rebalance every N
+    periods. This results in trading the factor at every period and the results
+    don't dependent on a spacific starting date.
+
     Parameters
     ----------
     factor_returns : pd.Series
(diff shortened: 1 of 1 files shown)


luca-s:
@twiecki Could you please double check the explanation I added is clear?

HereticSK:
If we are to mimic the cumulative return for the algorithm that holding N equal weighted sub-portfolios and rebalancing them every N period, I wonder if it would make more sense to compute the one-period return as `np.nanmean((1 + ret) ** (1. / period) - 1)`? 

Compared to the original algorithm, it computes geometric mean before arithmetic mean. For a given period, the one-period return for the whole portfolio equals the average of one-period returns for N sub-portfolios. On the other hand, assuming each sub-portfolio grows at a steady rate within each rebalancing cycle, the one-period return for each sub-portfolio is given by the geometric mean of the corresponding N-period return, i.e. `(1 + ret)**(1. / period) - 1)`. Combining these two steps, gives the one-period return of the whole portfolio equals `np.nanmean((1 + ret) ** (1. / period) - 1)`.

Moreover, in the original algorithm, `((np.nanmean(ret) + 1) ** (1. / period)) - 1`, the N-period returns for each sub-portfolio is averaged first. Somehow that seems like all these N-period returns are realized at this particular period, while they are actually realized at subsequent N periods. So I think it would make more sense to first compute (1 + ret) ** (1. / period) - 1), as an estimation of the realized return for a sub-portfolio on the given period, and then to average over all the sub-portfolios.

luca-s:
@HereticSK as this is a PR that only involves documenting the current functions behaviour, would you mind removing you last comment and copy paste it on the #187 issue? That would be much neat as we can keep discussing possible improvements on #187 issue, while this PR could be merged in the meanwhile as it only documents the behaviors that is already in the code

luca-s:
Closing this, I will provide another PR

PR #193, opened 2017-09-26
Improved utils.plot_events_distribution

luca-s:
The function it is now frequency agnostic (daily, minutely, etc data)
It is now possible to specify the resolution of the plot


twiecki:
Neat!

PR #198, opened 2017-10-02
Update version to 0.1.2

luca-s:
@twiecki time for a new version? Or too early?


twiecki:
Never too early.

luca-s:
I created a tag for the release

twiecki:
Can you add some brief release-notes to https://github.com/quantopian/alphalens/releases/tag/v0.1.2?

luca-s:
Sure, done.

twiecki:
Appreciate it!

luca-s:
@twiecki I just realized that Alphalens documentation says to install it via 'pip install git+https://github.com/quantopian/alphalens' but pypi has an old version of Alphalens too (v0.1.0). I think we should either update Alphalens release on pypi or remove that from pypi. What do you think?

Also, quantopian.github.io/alphalens still has the old v0.1.1 version.

twiecki:
Good catch, I just pushed a new version to pypi.

twiecki:
Also updated the docs. I think we should change the install instructions to `pip install alphalens` though.

luca-s:
Fine with me

PR #201, opened 2017-10-11
BLD: Use versioneer to pull version from git tag

richafrank:
Instead of having to update the code along with the git tag/release.

Be aware _version.py, versioneer.py, .gitattributes, and MANIFEST.in were all added/updated by versioneer itself, not manually by me.


luca-s:
LGTM

docs/source/conf.py
```diff
@@ -52,14 +54,8 @@
...
 
-# The version info for the project you're documenting, acts as replacement for
-# |version| and |release|, also used in various other places throughout the
-# built documents.
-#
-# The short X.Y version.
-version = u'0.1.2'
 # The full version, including alpha/beta/rc tags.
-release = u'0.1.2'
+release = version.split('+', 1)[0]
```
llllllllll: why strip this part?

richafrank: We do this for zipline, since we often deploy the docs from a commit slightly past the tagged release, where we might have fixed some docs issues but don't want to bother releasing a new version.

I guess the question is for @luca-s and @twiecki . Does this situation happen for alphalens and/or would you prefer to have the commit hash included where the docs display the version, e.g. something like `0.1.2+0.gdcb8814`?

twiecki: I think this is definitely useful.

richafrank: Ok, added back the commit hash.

PR #204, opened 2017-10-28
Improved "binning by group" option docstrings

luca-s:
The current meaning of 'by_group' option
(untils.get_clean_factor_and_forward_returns functions)  is a little obscure so the docstring was improved.

Also the name 'by_group' is used elsewhere in the API with the meaning of breaking the output in multiple results, one per group,so  the variable was renamed  to 'binning_by_group'

### DIFF
--- a/alphalens/utils.py
+++ b/alphalens/utils.py
@@ -294,7 +294,7 @@ def get_clean_factor_and_forward_returns(factor,
         entire time period of the passed factor data.
     binning_by_group : bool
         If True, compute bin or quantile buckets separately for each group.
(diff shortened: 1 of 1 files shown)


luca-s:
Closing this, I'll merge this change in another PR

luca-s:
@twiecki my only concern is that I renamed a function argument. Can we do that? Or should we keep the api stable? I believe we are still in the phase of API rfiniment, but I'll let you decide.

luca-s:
Included in PR #206

PR #206, opened 2017-10-29
Added support for group neutral factor analysis

luca-s:
I am working on a sector neutral factor and I discovered that Alphalens analysis on a group neutral factor is currently limited. With group neutral factor I mean a factor intended to rank stocks across the same group, so that it makes sense to compare performance of top vs bottom stocks in the same group but it doesn't make sense to compare performance of a stock in one group with performance of another group.

The main shortcoming is the return analysis: as there is no way to demean the returns by group,  the statistics shown are of little use. Also when the plots are broken down by group, the results are not useful either as there are 2 bugs:
- API documentation claims that perf.mean_return_by_quantile was performing is performing the demeaning at group level when splitting the plots by group, but even in that case it is not true
- The same goes for perf.mean_information_coefficient

Also changed the API arguments names in a consistent way:
- the tears.* functions use 'long_short' and 'group_neutral' all around, as those functions are intended to be the top level ones. 'by_group' is used if the function support multiple plots output, one for each group
- the remaining API (mostry performance.*) use 'demeaned' and  'group_adjust' as they used to be

### DIFF
--- a/alphalens/performance.py
+++ b/alphalens/performance.py
@@ -60,12 +60,12 @@ def factor_information_coefficient(factor_data,
     factor_data = factor_data.copy()
 
     grouper = [factor_data.index.get_level_values('date')]
-    if by_group:
-        grouper.append('group')
 
     if group_adjust:
         factor_data = utils.demean_forward_returns(factor_data,
                                                    grouper + ['group'])
+    if by_group:
+        grouper.append('group')
 
     ic = factor_data.groupby(grouper).apply(src_ic)
     ic.columns = pd.Int64Index(ic.columns)
@@ -127,7 +127,7 @@ def mean_information_coefficient(factor_data,
     return ic
 
 
-def factor_returns(factor_data, long_short=True, group_neutral=False):
+def factor_returns(factor_data, demeaned=True, group_adjust=False):
     """
     Computes period wise returns for portfolio weighted by factor
     values. Weights are computed by demeaning factors and dividing
@@ -141,13 +141,14 @@ def factor_returns(factor_data, long_short=True, group_neutral=False):
         each period, the factor quantile/bin that factor value belongs to, and
         (optionally) the group the asset belongs to.
         - See full explanation in utils.get_clean_factor_and_forward_returns
-    long_short : bool
-        Should this computation happen on a long short portfolio? if so, then
-        factor values will be demeaned across factor universe when factor
-        weighting the portfolio.
-    group_neutral : bool
-        If True, compute group neutral returns: each group will weight
-        the same and returns demeaning will occur on the group level.
+    demeaned : bool
+        Should this computation happen on a long short portfolio? if True,
+        then factor values will be demeaned across factor universe when
+        factor weighting the portfolio.
+    group_adjust : bool
+        Should this computation happen on a group neutral portfolio? If True,
+        compute group neutral returns: each group will weight the same and
+        returns demeaning will occur on the group level.
 
     Returns
     -------
@@ -164,13 +165,13 @@ def factor_returns(factor_data, long_short=True, group_neutral=False):
             return group / group.abs().sum()
 
     grouper = [factor_data.index.get_level_values('date')]
-    if group_neutral:
+    if group_adjust:
         grouper.append('group')
 
     weights = factor_data.groupby(grouper)['factor'] \
-        .apply(to_weights, long_short)
+        .apply(to_weights, demeaned)
 
-    if group_neutral:
+    if group_adjust:
         weights = weights.groupby(level='date').apply(to_weights, False)
 
     weighted_returns = \
@@ -182,7 +183,7 @@ def factor_returns(factor_data, long_short=True, group_neutral=False):
     return returns
 
 
-def factor_alpha_beta(factor_data, long_short=True):
+def factor_alpha_beta(factor_data, demeaned=True, group_adjust=False):
     """
     Compute the alpha (excess returns), alpha t-stat (alpha significance),
     and beta (market exposure) of a factor. A regression is run with
@@ -198,10 +199,12 @@ def factor_alpha_beta(factor_data, long_short=True):
         each period, the factor quantile/bin that factor value belongs to, and
         (optionally) the group the asset belongs to.
         - See full explanation in utils.get_clean_factor_and_forward_returns
-    long_short : bool
-        Should this computation happen on a long short portfolio? if so, then
-        factor values will be demeaned across factor universe when factor
-        weighting the portfolio.
+    demeaned : bool
+        Control how to build factor returns used for alpha/beta computation
+        -- see performance.factor_return for a full explanation
+    group_adjust : bool
+        Control how to build factor returns used for alpha/beta computation
+        -- see performance.factor_return for a full explanation
 
     Returns
     -------
@@ -210,7 +213,7 @@ def factor_alpha_beta(factor_data, long_short=True):
         for the given factor and forward returns.
     """
 
-    returns = factor_returns(factor_data, long_short=long_short)
+    returns = factor_returns(factor_data, demeaned, group_adjust)
 
     universe_ret = factor_data.groupby(level='date')[
         utils.get_forward_returns_columns(factor_data.columns)] \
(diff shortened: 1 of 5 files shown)


luca-s:
In  alphalens/tests/test_tears.py I removed few parameterized tests from:
- test_create_summary_tear_sheet
- test_create_turnover_tear_sheet
- test_create_information_tear_sheet
- test_create_returns_tear_sheet

Those fuctions are called by test_create_full_tear_sheet anyway, so no need to be paranoic

luca-s:
@twiecki I also think that the information analysis doesn't need the group_adjust (now group_neutral) flag. 

Currently and historically that flag is used to perform returns demeaning before computing the IC between factor values and returns. Subtracting a constant (the mean value) to the returns doesn't affect the value of IC (IC is Spearman's rank correlation coefficient), so we should get rid of this flag all around the information analysis API, unless there is a plan to change the way IC is computed (but we can still restore the flag in the future if needed).

Please let me know what you think and in case I'll add a commit in this PR to remove the flag.

luca-s:
I couldn't resist, I removed the group_adjust/group_neutral flags from  information analysis API, but I didn't
squash the commit so it's a change on its own

alphalens/performance.py
```diff
@@ -60,21 +57,17 @@ def src_ic(group):
     factor_data = factor_data.copy()
 
     grouper = [factor_data.index.get_level_values('date')]
+
     if by_group:
         grouper.append('group')
 
-    if group_adjust:
-        factor_data = utils.demean_forward_returns(factor_data,
```
twiecki: I'm confused, these aren't the ICs, right? So de-meaning should make a difference here.

twiecki: Oh I get it, it doesn't matter for computing the IC which is what we're doing here.

alphalens/performance.py
```diff
@@ -407,20 +406,25 @@ def factor_rank_autocorrelation(factor_data, period=1):
     return autocorr
 
 
-def average_cumulative_return_by_quantile(quantized_factor,
+def average_cumulative_return_by_quantile(factor_data,
```
twiecki: Why this change?

alphalens/performance.py
```diff
@@ -432,20 +436,82 @@ def average_cumulative_return_by_quantile(quantized_factor,
...
+                                          True, True, demean_by)
+
+    def average_cumulative_return(q_fact, demean_by):
+        q_returns = cumulative_return(q_fact, demean_by)
         return pd.DataFrame({'mean': q_returns.mean(axis=1),
                              'std': q_returns.std(axis=1)}).T
 
-    return quantized_factor.groupby(quantized_factor) \
-        .apply(average_cumulative_return)
+    if by_group:
```
twiecki: This code could use some comments.

alphalens/performance.py
luca-s: yes, if apply  utils.demean_forward_returns to the returns the resulting IC is the same as not applying utils.demean_forward_returns

twiecki:
This looks great! I think the code could use a few more comments in general and I wonder which backwards compatibility we're breaking here and if we should check for that.

alphalens/performance.py
luca-s: I usually comment a lots but because Alphalens don't have comments at all I have almost always refrained from adding them, I thought it was kind of a policy :D

twiecki: Haha, not at all :)

luca-s:
Yeah, we break few APIs here, let me list all the changes

luca-s:
On a high level view, the new API work as before if the user was running Alhalens with default values.
The only scenario where the changes have an impact is if the user was using the by_group options to view plots by group

Main user API brekage:
- tears.create_summary_tear_sheet: added optional group_neutral (no breakage)
- tears.create_returns_tear_sheet:  added optional group_neutral (brekage because it's not the last argument)
- tears.create_information_tear_sheet:  removed group_adjust (brekage)
- tears.create_full_tear_sheet: renamed group_adjust->group_neutral (brekage)
- tears.create_event_returns_tear_sheet:  added optional group_neutral (brekage because it's not the last argument)


Lower level API breakage:
- utils.get_clean_factor_and_forward_returns:  renamed by_group->binning_by_group
- perf.factor_information_coefficient: removed group_adjust
- perf.mean_information_coefficient: removed group_adjust
- perf.factor_returns: rename long_short->demeaned, renamed group_neutral->group_adjust
- perf.factor_alpha_beta: renamed long_short->demeaned, added optional group_adjust
- perf.mean_return_by_quantile: added group_adjust
-perf.average_cumulative_return_by_quantile: added group_adjust, added by_group, changed quantized_factor->factor_data

twiecki:
I think we should add deprecation warnings for our changes in the high-level API. For example, for renaming a kwarg, we should check the old name, raise the warning, but internally just remap to the new name.

luca-s:
There is a simpler solution if we prefer backward compatibility over nicer arguments name: 
- there will be no breakage in create_returns_tear_sheet and create_event_returns_tear_sheet if the new argument 'group_neutral' is added as last argument (now is the last but one because I liked to have it close to long_short as they are very related)
-  We keep the name group_adjust instead of group_neutral

luca-s:
Actually we can also leave the argument 'group_adjust' to tears.create_information_tear_sheet and document it as deprecated. It doesn't affect the computation so we keep it only to avoid API change. 
This and all the above, we can avoid API breakage

twiecki:
I wouldn't shy away from breaking API if it makes things easier to understand (which I think this does). And the deprecation warnings should be pretty straight-forward to implement so that would be my preference.

luca-s:
Good, I'll go for  deprecation warnings then

twiecki:
This looks great, as this adds API change, we should get Quantopian engineering to take a look. CC @richafrank.

twiecki:
```
alphalens/tears.py:82:25: E128 continuation line under-indented for visual indent

alphalens/tears.py:92:13: W291 trailing whitespace

alphalens/tears.py:119:1: W293 blank line contains whitespace

alphalens/tears.py:124:13: W291 trailing whitespace

alphalens/tears.py:139:12: E111 indentation is not a multiple of four

alphalens/tears.py:140:23: E128 continuation line under-indented for visual indent

alphalens/tears.py:141:23: E128 continuation line under-indented for visual indent

alphalens/tears.py:142:23: E128 continuation line under-indented for visual indent

alphalens/tears.py:143:23: E128 continuation line under-indented for visual indent

alphalens/tears.py:152:13: W291 trailing whitespace

alphalens/tears.py:171:12: E111 indentation is not a multiple of four

alphalens/tears.py:172:16: E128 continuation line under-indented for visual indent

alphalens/tears.py:173:16: E128 continuation line under-indented for visual indent

alphalens/tears.py:174:16: E128 continuation line under-indented for visual indent

alphalens/tears.py:175:16: E128 continuation line under-indented for visual indent

alphalens/tears.py:175:58: W291 trailing whitespace
```

luca-s:
@richafrank above you can find the list of API breakage. The last commit add backward compatibility for the main functions, the ones in tears.*:
- functions call that use keyword arguments are fully backward compatible
- function calls that use positional arguments can be safe most of the time, but for those scenarios (and only on those) where the call is ambiguous and a mapping to the new API was not possible, I made sure to raise a warning.

richafrank:
Thanks @luca-s . I think quantopian's internal apps don't depend on these interfaces, just users' notebooks.

@abhijeetkalyan This means that updating the version of alphalens could mean users need to update their notebooks. I assume that's fine, right?

abhijeetkalyan:
@richafrank @luca-s it sounds like with https://github.com/quantopian/alphalens/pull/206/commits/362612cf32abde6367b2b789f526ce7a7bb0fceb, this PR is fully backwards compatible. Is that correct, or is there still any API breakage from https://github.com/quantopian/alphalens/pull/206#issuecomment-341097404 that would occur?

luca-s:
@abhijeetkalyan two scenarios are still not backward compatible and a warning is raised in that case:
- tears.create_returns_tear_sheet when run with exactly 3 positional arguments
- tears.create_event_returns_tear_sheet when run with exactly 5 positional arguments

luca-s:
I'd like to go on and merge this PR. The scenarios where the API are not backward compatible are very limited in scope and and they are confined to the cases where the users wanted to break down the plots by group(e.g. sector) AND they didn't provide keyword arguments. Even if those scenarios are broken by this API change Alphalens will raise a warning and the output won't provide aplot for each group, making evident that something went wrong and so encouraging the users to look at the warning.

twiecki:
@luca-s Yep, seems like we're all set. Thanks!

PR #209, opened 2017-10-31
DOC: updated installation instruction

luca-s:
@twiecki  As agreed [here](https://github.com/quantopian/alphalens/pull/198) the preferred way to install Alphalens is now from pypi instead of gitub


twiecki:
LGTM, thanks!

PR #210, opened 2017-11-01
max_loss increased to 25% by default

luca-s:
Don't be too strict with max_loss default value, that would be frustrating. So print the amount of dropped factor data even when the limit is not reached, so that the user is aware of what happened. 

Also the code is nicer now, the Exceptions that are suppressed by utils.quantize_factor can be now seen setting max_loss=0 so that troubleshooting potential issues is still possible.


twiecki:
Nice!

PR #214, opened 2017-11-02
Changed pd.rolling_mean() to use the new *.rolling().mean() API

mmargenot:
I changed the instances of `rolling_mean` to use the new API. The instances of `rolling_apply` that are used elsewhere should be updated on top of these.

### DIFF
--- a/alphalens/plotting.py
+++ b/alphalens/plotting.py
@@ -523,7 +523,7 @@ def plot_mean_quantile_returns_spread_time_series(mean_returns_spread,
     mean_returns_spread_bps = mean_returns_spread * DECIMAL_TO_BPS
 
     mean_returns_spread_bps.plot(alpha=0.4, ax=ax, lw=0.7, color='forestgreen')
-    mean_returns_spread_bps.rolling(window=22).plot(
+    mean_returns_spread_bps.rolling(window=22).mean().plot(
         color='orangered',
         alpha=0.7,
         ax=ax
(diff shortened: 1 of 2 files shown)


luca-s:
LGTM, as long as @twiecki agree with the new requirement "pandas>=0.18.0"

luca-s:
@mmargenot as it is related to this commit,  would you remove the rolling_apply inside utils.py as well? 

Current code
```
    sub_portfolios = pd.rolling_apply(sub_portfolios,
                                      period,
                                      rate_of_returns,
                                      min_periods=1,
                                      args=(period,))   
```
That should become:
```
    sub_portfolios = sub_portfolios.rolling(window=period, min_periods=1) \
                                   .apply(rate_of_returns, args=(period,))         
```

twiecki:
Thanks @mmargenot, changing the requirement to >= 0.18 is fine.

twiecki:
CC @richafrank.

richafrank:
Thanks. The requirement bump to 0.18 sounds good to me, as that matches the lower bound for [zipline](https://github.com/quantopian/zipline/blob/master/etc/requirements.txt#L20). 

I assume the pandas API change here is in fact supported by 0.18? Confirming only because the travis build is testing with latest pandas (0.20).

twiecki:
Yeah, I was worried about that too but thought I misremembered. I don't think 0.18 supports `.rolling().mean()` yet.

Should probably also add a test-matrix for older pandas versions here.

mmargenot:
The API reference for pandas has `New in version 0.18.0.` for both `pandas.Series.rolling` and `pandas.DataFrame.rolling`, [here](https://pandas.pydata.org/pandas-docs/version/0.18.1/generated/pandas.Series.rolling.html) and [here](https://pandas.pydata.org/pandas-docs/version/0.18.1/generated/pandas.DataFrame.rolling.html). I can make it 0.18.1 just to be safe?

mmargenot:
@luca-s I can definitely fix the `rolling_apply`. I just wanted to get this one first.

twiecki:
Thanks @mmargenot!

Issue #225, opened 2017-11-19
Integration with pyfolio and Quantopian's new Risk Model


luca-s:
I have been thinking about the nice "Alpha decomposition" #180 feature. The information it provides is actually a small part of what already available in pyfolio  and the Quantopian's new Risk Model. On one side we cannot replicate all the information provided by those two tools but on the other side it would be great to have all that analysis without having to build an algorithm and run a backtest, something that could be integrated into Alphalens.

Then, why don't we create a function in Alphalens that builds the input required by  pyfolio  and the Quantopian's new Risk Model? Alphalens already simulates the cumulative returns of a portfolio weighted by factor values, so we only need to format those information in a way that is compatible with the other two tools. That would be a pure theoretical analysis, but except for commissions and slippage the results would be realistic and it also would serve as benchmark for the algorithm (users can compare the algorithm results, after setting commission and slippage to 0, with these theoretical results and check if they implemented the algorithm correctly).

I haven't looked at pyfolio in details so I don't know the details of the input, but if @twiecki can help me with those details I can work on this feature and the same for  Quantopian's new Risk Mode (I don't know if that is part of pyfolio or a separate project).

twiecki:
That's really clever!

Pyfolio just requires returns and positions. I think we should be able to
get both of those from alphalens pretty easily.

luca-s:
@twiecki just need to double check with you few details about pyfolio input:

From API documentaion:

 ```
   returns : pd.Series
        Daily returns of the strategy, noncumulative.
         - Time series with decimal returns.
         - Example:
            2015-07-16    -0.012143
            2015-07-17    0.045350
            2015-07-20    0.030957
            2015-07-21    0.004902
```
Do the `returns` need to be computed at the end of the trading day? For example, does the entry "2015-07-16    -0.012143" mean the portfolio had a return of  -0.012143 at market close on 2015-07-16?



```
    positions : pd.DataFrame, optional
        Daily net position values.
         - Time series of dollar amount invested in each position and cash.
         - Days where stocks are not held can be represented by 0 or NaN.
         - Non-working capital is labelled 'cash'
         - Example:
            index         'AAPL'         'MSFT'          cash
            2004-01-09    13939.3800     -14012.9930     711.5585
            2004-01-12    14492.6300     -14624.8700     27.1821
            2004-01-13    -13853.2800    13653.6400      -43.6375 
```
Do the `positions` need to be computed at the end of the trading day too? Or beginning of trading day?

[...]

PR #227, opened 2017-12-07
ENH Pyfolio integration

luca-s:
Issue #225 

Added performance.create_pyfolio_input, which create input suitable for pyfolio. Intended use:

```
factor_data = alphales.utils.get_clean_factor_and_forward_returns(factor, prices)

pf_returns = alphales.performance.create_pyfolio_input(factor_data,'1D',
                     long_short=True,
                     group_neutral=False,
                     quantiles=None,
                     groups=None)

pyfolio.tears.create_returns_tear_sheet(pf_returns)
```

Also, I  greatly improved the function that computes the cumulative returns as that is the base on which the pyfolio integration feature is built on.  Mainly I removed the legacy assumptions  which required the factor to be computed at specific frequency (daily, weekly etc) and also the periods had to be multiple of this frequency (2 days, 4 days etc). The function is now able to properly compute returns when there are gaps in the factor data or when we analyze an intraday factor.


luca-s:
Oh, I forgot. To properly compute the portfolio returns I needed to know the length of the periods. Currently they are just integers and it is not known the units of those integers. So I changed the period column names from a simple integer to a pd.Timedelta string representation (1D, 5H, etc)

luca-s:
I'll go through pep8 issues tomorrow.

luca-s:
Ok, finished with pep8 issues. This PR is ready

alphalens/performance.py
```diff
@@ -355,8 +496,10 @@ def quantile_turnover(quantile_factor, quantile, period=1):
         DataFrame with date, asset and factor quantile.
     quantile : int
         Quantile on which to perform turnover analysis.
-    period: int, optional
+    period: pandas.Timedelta or string
         Period over which to calculate the turnover
+        if 'period' is a string it must follow pandas.Timedelta constructor
+        format (.e.g '1 days', '1D', '30m', '3h', '1D1h', etc)
```
twiecki: `.e.g` -> `e.g.`

twiecki:
This looks great!

luca-s:
I am pretty sure there are still many issues because I haven't done many tests yet, especially because I cannot access Quantopian data on my local machine. Anyway, let's consider this an experimental feature for now and then we will improve it.

luca-s:
I realized I wasn't clear in my previous comment. What I mean is that the PR is ready to be merged but I expect some issues with pyfolio as I haven't done extensive testing. Anyway I don't expect regressions with existing functionalities and I actually expect better compatibility with non-daily factors and event studies.

luca-s:
@twiecki I am going to merge this if you agree

alphalens/utils.py
```diff
@@ -680,8 +640,36 @@ def std_conversion(period_std):
...
 def get_forward_returns_columns(columns):
-    return columns[columns.astype('str').str.isdigit()]
+    pattern = re.compile(r"^(\d+([Dhms]|ms|us|ns))+$", re.IGNORECASE)
+    valid_columns = [(pattern.match(col) is not None) for col in columns]
+    return columns[valid_columns]
+
+
+def timedelta_to_string(timedelta):
+    """
+    Utility that converts a pandas.Timedelta to a string rappresentation
```
twiecki: representation

alphalens/utils.py
```diff
@@ -680,8 +640,36 @@ def std_conversion(period_std):
...
 
 
 def get_forward_returns_columns(columns):
-    return columns[columns.astype('str').str.isdigit()]
+    pattern = re.compile(r"^(\d+([Dhms]|ms|us|ns))+$", re.IGNORECASE)
+    valid_columns = [(pattern.match(col) is not None) for col in columns]
+    return columns[valid_columns]
+
+
+def timedelta_to_string(timedelta):
```
twiecki: Does that really not exist inside pandas?

luca-s: Unfortunately no, the pandas version gives back the verbose version. e.g. '1 days 00:00:00'

alphalens/utils.py
luca-s: thanks :)

twiecki:
Does this change existing behavior?

luca-s:
No, all scenarios that were working before are working now + the current behaviour does a better job in handling factors with gaps (when values are not computed every day, e.g. event study) and intraday factors.

The only user level visible change is the period column names that are now named "1D", "3D" (1 day, 3 days or "1H", 3H"  for 1 hour, 2 hours, etc. ) instead of "1", "2".

twiecki:
OK, great. And users can now also have arbitrary periods, like hours.

I haven't reviewed the code carefully but I'm fine merging this as an experimental feature.

twiecki:
Actually, let's see if @richafrank has someone who could do a code review.

luca-s:
Actually arbitrary periods are possible even without this PR, for example a daily factor with 1 hours period or a weekly factor (computed on Monday for example) with a period of 3 days. With this kind of factors the results are computed correctly even though the turnover analysing is not so interesting, but the cumulative returns are wrong and this prevent the pyfolio analysis, this is why I fixed it and as I was already fixing the code I made sure it works with factors not  computed 
at a fixed frequency, e.g. event study

abhijeetkalyan:
@luca-s @twiecki this looks neat! I had a couple questions around the UX:
- The name confused me a bit. When I saw `create_pyfolio_input`, I expected it to have positions and transactions as well. Since the function's output is the return stream, should we change its name to clarify that it's providing only returns, and not all of the other inputs that pyfolio consumes?
- If this is still an experimental feature as @twiecki mentioned in https://github.com/quantopian/alphalens/pull/227#issuecomment-351047163, should we add some kind of warning to inform users of that fact?

luca-s:
You are right @abhijeetkalyan , the name is confusing a little. The reason is that I initially thought to return  `positions` and `transactions` as well but I haven't managed to do that yet. I am actively working on this so my plan is to improve `create_pyfolio_input` to make it returns `positions` and `transactions`  in another PR.  I hope to push that PR soon together with an example NB. If I manage to do that before we create a new tag I would avoid the warning, otherwise if we create a tag and release a new version of Alphalens before I finish working on this feature I will add a warning message. Sounds like a plan?

luca-s:
I added an example NB and also I made `create_pyfolio_input` returns the benchmark returns too (as the factor universe mean returns).

luca-s:
I also added a warning in `create_pyfolio_input` docstring:

>  WARNING: this API is still in experimental phase and input/output
>                      paramenters might change in the future 

I believe this PR is now ready to be merged

alphalens/performance.py
```diff
@@ -68,7 +68,7 @@ def src_ic(group):
         grouper.append('group')
 
     ic = factor_data.groupby(grouper).apply(src_ic)
-    ic.columns = pd.Int64Index(ic.columns)
+    ic.columns = pd.Index(ic.columns)
```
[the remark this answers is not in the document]
luca-s: I guess you are right :)

alphalens/utils.py
```diff
@@ -604,66 +613,17 @@ def common_start_returns(factor,
...
     """
-    return period_ret.add(1).pow(1. / period_ret.name).sub(1)
+    period_len = period_ret.name
+    conversion_factor = (pd.Timedelta(one_period_len) /
+                         pd.Timedelta(period_len))
+    return period_ret.add(1).pow(conversion_factor).sub(1)
 
 
-def std_conversion(period_std):
+def std_conversion(period_std, one_period_len):
```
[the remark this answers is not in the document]
luca-s: you are right!

luca-s:
@prsutherland thanks for reminding me about the walk through NB I completely forgot about that. Regarding the  default values on the `period` parameters I am not sure... If I find a sensible default I would restore them.
The movement from integers to Timedelta strings for period was needed to carry information that otherwise would have been lost but some thorough testing is now required because this change could lead to some hidden and non-obvious bugs. Sometimes I think get_clean_factor_and_forward_returns should return an object  where we can store all the information we want together with factor_data instead of trying to save the information we need inside the DataFrame column names...but for now let's keep it this way.

luca-s:
Unless there are other comments I am going to merge this PR soon.  I only need to finish updating the tear_sheet_walk_through NB.

twiecki:
Looks great, @luca-s. Thanks!

PR #245, opened 2018-01-05
ENH: Changed palette to work with colorblind users

mmargenot:


luca-s:
@mmargenot is your idea to remove hard coded G/R palettes and use matplotlib default? I like that but there are two issues:
- two places still have hard coded values (G/R palette): [here](https://github.com/quantopian/alphalens/blob/master/alphalens/plotting.py#L700) and [here](https://github.com/quantopian/alphalens/blob/master/alphalens/plotting.py#L779) 
- the plots where all quantiles are shown at the same time can be easier to read with diverging color palette. So I would still hard code a diverging color palette in both [here](https://github.com/quantopian/alphalens/blob/master/alphalens/plotting.py#L779) and [here](https://github.com/quantopian/alphalens/blob/master/alphalens/plotting.py#L826) but I would change the G/R palette to something else.

luca-s:
uhmmm, I thought again at that and I believe the three places where the palette is hard coded should stay the same as all of them need a diverging color palette. Easiest thing is to just replace the palette to something else. What is your view @mmargenot ?

mmargenot:
I can change the hardcoded spots to use a more friendly diverging color palette. I think I like cool/warm since it looks like the smoothest transition.

Issue #246, opened 2018-01-07
Migrate tests from deprecated nose-parameterized


jameschristopher:
`nose-parameterized` has been deprecated, renamed, and generalized. It now lives as just `parameterized` See [Migration Guide](https://github.com/wolever/parameterized#migrating-from-nose-parameterized-to-parameterized)

twiecki:
Thanks for flagging this @jameschristopher!

PR #251, opened 2018-01-09
MAINT: Migrate tests from deprecated nose-parameterized

luca-s:
issue #246


twiecki:
Don't we need to change `requirements` and `.travis`?

luca-s:
I believe this is the only thing we have to do. Old nose_parameterized versions will fail but this is a developers issue (they need to update the module in case they are using an old one).

luca-s:
The error are timeouts. Since we added the tests for multiple pandas versions Travis fails very often. It might be the increase on the load?

luca-s:
By the way, if we don't want to apply this PR now it's fine. Travis will keep working and no developer will be forced to update nose_parameterized. Eventually, when the deprecation warning become and error we'll update the tests with this PR.

luca-s:
@twiecki you are right, even the module name was renamed to parameterized from nose_parameterized

twiecki:
Also here: https://github.com/quantopian/alphalens/blob/master/setup.py#L25

luca-s:
Fixed

twiecki:
Thanks!

jameschristopher:
Awesome Luca!

Issue #219, opened 2017-11-07
Interaction and Correlation Features


MichaelJMath:
This post describes a possible enhancement, but I'm not sure it fits in with the alphalens project objective as it deals with factor combination. I am proposing it because alphalens already has some of the functions already in place to make this type of analysis easy. However, I realize it may be better included in a separate project.

When combining alpha factors, analysts often look at the correlation between factors to determine whether we are adding anything "new" to the signal or if the two factors simply measure the same thing.
(Ideally we would want factors with low correlation).

Additionally, it is interesting to know whether two factors might have an interaction effect where instead of the effects combining in an additive manner, they do so in a multiplicative manner.

References on Interaction Effects:
https://courses.washington.edu/smartpsy/interactions.htm
https://en.wikipedia.org/wiki/Interaction_(statistics

Below, I have included a link to a Jupyter notebook that uses the size and momentum factors from the [Ken French data library](http://mba.tuck.dartmouth.edu/pages/faculty/ken.french/data_library.html) as an example of what the output might look like. (The implementation in alphalens if included would be different of course). https://www.dropbox.com/s/owh8rfumuch4c9g/Momentum%20and%20Size%20Interaction-2.ipynb?dl=0

Here is a snippet of the output from the notebook:
![heatmap](https://user-images.githubusercontent.com/28760724/32498468-cbe12d62-c39d-11e7-80e7-abeb63002d15.PNG)
![interaction plot](https://user-images.githubusercontent.com/28760724/32498473-cfff63dc-c39d-11e7-9896-46321103d12a.PNG)
![correlation plot](https://user-images.githubusercontent.com/28760724/32498481-d480fbaa-c39d-11e7-9545-f55c3b378719.PNG)

As mentioned above, I'm not 100% sure this fits in with the alphalens project, as alphalens tends to focus on the analysis of a single alpha factor. However, the reason I bring it up is that alphalens already has some of the functions that make this type of factor combination analysis easy. Essentially, the factor_data dataframe (returned from `utils.get_clean_factor_and_forward_returns()`) contains the data needed for the interaction analysis (forward returns, factor values and quantiles).  

Any feedback or questions are appreciated.

luca-s:
I see this inter-factor analysis as complementary to Alphalens and I believe it would fit well in the project. I thought about this feature myself few times, even though I didn't think about the specific information I wanted to see.

[...]

Issue #253, opened 2018-01-10
Feature request: Use alphalens with returns instead of prices


neman018:
Is there a way to run Alphalens using my own custom input matrix of returns (ie. custom factor adjusted) rather than inputting Prices (ie. the "pricing" table in the examples) for each period?

luca-s:
You could skip the call to `utils.get_clean_factor_and_forward_returns` and format your data accordingly to what the tears.* functions expect (the format is described in the return value of `utils.get_clean_factor_and_forward_returns` ). 

More generally `utils.get_clean_factor_and_forward_returns`  is an utility function, you don't have to call it if you don't need it but you can have a look at the source code, that might help you formatting your data .

luca-s:
@all
I wonder how often it happens that the returns are available instead of prices. If that's a common scenario we have to think about providing an utility function that prepare and format the input data from returns information.

neman018:
Any chance you guys will add a utility function for returns (vs stock prices)? I think this might be useful for a lot users who construct their own factor models and consequently custom return streams

HereticSK:
I believe it is quite common to iterate different factors while returns do not have to change. It would be a lot easier and faster to have a utility function that construct input data from returns. 

Currently, I implement this on my own. But I believe it may be a good idea to include this in the package, if it is common enough.

twiecki:
@luca-s Seems like there is indeed quite a bit demand for it.

HereticSK:
Actually we've discussed something similar in #199 . I think the pain point here is that `get_clean_factor_and_forward_returns` repeats the slow forward return computation every time.  So I simply split out this part and reuse everything else.

I make this `create_factor_data` function based on `get_clean_factor_and_forward_returns`, where computing forward returns is replaced with pre-computed values, and in the signature `price` is replaced with `forward_returns`, `periods ` removed. 
```Python
def create_factor_data(factor,
                       forward_returns,
                       groupby=None,
                       by_group=False,
                       quantiles=5,
                       bins=None,
                       groupby_labels=None):

    # Replace calling compute_forward_returns with pre-computed values
    merged_data = forward_returns.copy() 

    # Keep everything else unchanged below
    ......
```

Now the analysis code becomes something like:
```
periods = (20, 40, 60, 120)
forward_returns = alphalens.utils.compute_forward_returns(prices=pricing, periods=periods)

factor_data_1 = create_factor_data(
    factor=factor_1, 
    forward_returns=forward_returns,
    quantiles=10
[...]

Issue #218, opened 2017-11-06
Test alphalens.tears.* functions: add matplotlib output comparison


luca-s:
Currently the tests relative to alphalens.tears.* functions check that no exception are thrown only. It would be nice to capture matplotlib output, compute a checksum for each figure generated and save those as reference for future tests. That would be a much robust test than what is currently performed

luca-s:
In order to accomplish this we need to force a specific matplotlib version for Travis CI, otherwise small differences in the plots due to matplotlib version discrepancies would result in test failures. 

Does anybody see a problem in forcing a matplotlib version?

richafrank:
👍  for reproducible results

PR #258, opened 2018-01-28
Two Factor Interaction Development: Initial Data Structure, Test Module, and Plot

MichaelJMath:
This begins the development of the "factors_interaction_tearsheet" from issue #219. The goal of this pull request is to get feedback on whether this branch seems to be going in the right direction.

**Description of Changes**
1. Create `join_factor_with_factor_and_forward_returns` function
    - Creates a function complimentary to `get_clean_factor_and_forward_returns` that joins an additional factor to the `factor_data` dataframe returned by `get_clean_factor_and_forward_returns`. 
    - This new dataframe returned, call it "`multi_factor_data`", will be the core source/data structure providing the necessary data for the factors_interaction_tearsheet computations.
2.  Create an associated test module.
3. Modify `perf.mean_return_by_quantile` to take an additional parameter so that it can group by multiple factor quantiles.
4. Add first plotting function, `plot_multi_factor_quantile_returns`, to create an annotated heatmap of mean returns by two-factor quantile bins.
5. Create the `tears.create_factors_interaction_tear_sheet` as the entry point to the multi-factor tearsheet.

**Requesting Feedback**
1. Comments and suggestions on the `utils.join_factor_with_factor_and_forward_returns` function
     1. Should there be a wrapper that builds the multi_factor_data dataframe in one step. (i.e. wrap this function with `get_clean_factor_and_forward_returns`?
2. I'm not too familiar with creating effective unit tests, so any feedback on this module is appreciated.
3. In regards to Change 3 above:
    1. My first thought, following suggestion of @luca-s, was to create a separate performance module which would contain all functions for this sort of computation.
    2. Since the existing performance module already contains a lot of the needed functionality, I thought maybe I would create a wrapper function in this new module that added the necessary functionality.
    3. However, in `perf.mean_return_by_quantile`, I needed to add a parameter to this function to make it work in a clean manner. Not sure how I could have done that with a wrapper.
    4. So I guess my question is, what are the community's thoughts on how I dealt with this particular issue, and also what are thoughts on related considerations going forward?
4. Any other comments/guidance on path of development going forward is greatly appreciated.
5. Also, let me know if there are too many changes in this pull request for efficient/easy review.


luca-s:
@MichaelJMath there seems to be a lot of development here, thank you very much for your work.  I will start reviewing this PR soon. Would you mind rebasing your branch on quantopian:factors_interactions? I have just rebased it on quantopian:master.

MichaelJMath:
@luca-s I tried rebasing this morning, and git said everything was up-to-date. I think I mistakenly created the new branch off master instead of factors_interactions.

Would it be easier for me to just close this pull request and open up a duplicate one now that factors_interaction has been rebased on master? I'm assuming this would remove all of the old commits showing on this pull request that are unrelated to this request.

luca-s:
I believe your solution would work but the scenario we are facing it is a very common one so it should be pretty easy to fix without creating a new PR:

```
# update your upstream information
git fetch upstream -p 

# rebase on top of quantopian:factors_interactions
git checkout  two_factor_interaction
git rebase upstream/factors_interactions
git push origin two_factor_interaction
```

MichaelJMath:
I tried the following at my Bash prompt. I'm still troubleshooting, but figured I'd post my progress.

```
Mike (factors_interactions) alphalens $ git remote -v
    origin  https://github.com/MichaelJMath/alphalens.git (fetch)
    origin  https://github.com/MichaelJMath/alphalens.git (push)
    upstream        https://github.com/quantopian/alphalens.git (fetch)
    upstream        https://github.com/quantopian/alphalens.git (push)

Mike (factors_interactions) alphalens $ git fetch upstream -p
Mike (factors_interactions) alphalens $ git checkout two_factor_interaction
    Switched to branch 'two_factor_interaction'

Mike (two_factor_interaction) alphalens $ git rebase upstream/factors_interactions
    Current branch two_factor_interaction is up to date.

Mike (two_factor_interaction) alphalens $ git push origin two_factor_interaction
    Everything up-to-date
```

luca-s:
So you were right, your branch `two_factor_interaction` is up-to-date. Could you try pushing with --force option? That might force github to refresh this PR. Otherwise you can go with your initial suggestion of opening a new PR.

twiecki:
This looks really promising @MichaelJMath!

MichaelJMath:
Same results for git push --force
```
Mike (two_factor_interaction) alphalens $ git push origin two_factor_interaction --force
    Everything up-to-date
```
If I look at my [network graph on Github](https://github.com/MichaelJMath/alphalens/network), it is clearly not showing the rebase as my two_factor_interaction branch is still based off master. 
mjm_network

This is in contrast to the [alphalens network graph](https://github.com/quantopian/alphalens/network), which is showing my branch based off factors_interactions.
alphalens_network

MichaelJMath:
I'm closing this PR and opening a duplicate to try and resolve the issue.

luca-s:
@MichaelJMath since `quantopian:factors_interactions` is a development branch I rebased/merged your PR there. Now it looks fine and we can all work on this branch.

luca-s:
@MichaelJMath I had a quick view at the PR, I like your change overall and here are my comments:

> 1. Comments and suggestions on the `utils.join_factor_with_factor_and_forward_returns` function
>      1. Should there be a wrapper that builds the multi_factor_data dataframe in one step. (i.e. wrap this function with `get_clean_factor_and_forward_returns`?

My idea was that the user have to call `get_clean_factor_and_forward_returns` on both factor1 and factor2 data to obtain factor_data1 and factor_data2 and then to call `tears.create_factors_interaction_tear_sheet` passing the latter two. Is there an advantage to merge the two factors data together? If so we could call `utils.join_factor_with_factor_and_forward_returns`  inside `tears.create_factors_interaction_tear_sheet` so that the users can be unaware of this detail.

Please let me know your thought on this.

> 2. I'm not too familiar with creating effective unit tests, so any feedback on this module is appreciated.

Your test looks fine. We haven't created a framework for testing tear sheets output yet (see PR #218) so testing only for exceptions is fine now. More generally speaking, there is a test for each functions that performs computation (so the functions inside performance and utils) as you haven't added new computational functions there is no need for more tests. Also it is ok starting with few tests that cover the most common scenarios and  then, once bugs are found, we can add more tests to avoid regressions.

As a suggestion, it's better to add tests at the end of the PR when the new API/functionalities are well defined, otherwise you might end up rewriting the tests.

> 3. In regards to Change 3 above:
>     1. My first thought, following suggestion of @luca-s, was to create a separate performance module which would contain all functions for this sort of computation.
>     2. Since the existing performance module already contains a lot of the needed functionality, I thought maybe I would create a wrapper function in this new module that added the necessary functionality.
>     3. However, in `perf.mean_return_by_quantile`, I needed to add a parameter to this function to make it work in a clean manner. Not sure how I could have done that with a wrapper.
>     4. So I guess my question is, what are the community's thoughts on how I dealt with this particular issue, and also what are thoughts on related considerations going forward?

I believe your change is fine. There are no general rules, depending on the particular case it might be better to add a wrapper or to add an argument, like you did.

We have to pay particular attention to the API primarily intended for the users (all the tears.* functions and utils.get_clean_factor_and_forward_returns). We try our best to not break them but if we do then we have to document it in the release notes.

> 4. Any other comments/guidance on path of development going forward is greatly appreciated.

The more you become familiar with the project the easier will be but I believe you already pay lots of attention to the code you write and I am happy with your PR. 

> 5. Also, let me know if there are too many changes in this pull request for efficient/easy review.

There is no limits in the amount of changes, but the idea is that you create commits so that who looks at the project history has a clue on what happened. They must summarize each new achievement in the project, a self-contained task, you don't want to enter too many details. Try to describe what you achieved with a commit and not what you did, they should look like a release note.

Also you can use one of this prefixes in the commit message:

> API: an (incompatible) API change
> BLD: change related to building numpy
> BUG: bug fix
> DEP: deprecate something, or remove a deprecated object
> DEV: development tool or utility
> DOC: documentation
> ENH: enhancement
> MAINT: maintenance commit (refactoring, typos, etc.)
> REV: revert an earlier commit
> STY: style fix (whitespace, PEP8)
> TST: addition or modification of tests
> REL: related to releasing numpy

luca-s:
I forgot two small details:

You don't want to call plt.show() [here](https://github.com/quantopian/alphalens/blob/factors_interactions/alphalens/plotting.py#L958) and it would be better to follow the example of the other tears.* functions and to create a  GridFigure in `create_factors_interaction_tear_sheet`, pass it to `plot_multi_factor_quantile_returns` and finally call  `plt.show()` and ` gf.close()` at the end of `create_factors_interaction_tear_sheet`.

Also, [this](https://github.com/quantopian/alphalens/blob/factors_interactions/alphalens/plotting.py#L950) you shouldn't rely on the active figure, it would be better to operate on the passed `ax`

MichaelJMath:
> My idea was that the user have to call get_clean_factor_and_forward_returns on both factor1 and factor2 data to obtain factor_data1 and factor_data2 and then to call tears.create_factors_interaction_tear_sheet passing the latter two.

On this point, I agree with allowing the user to pass factor1 and factor2 data into `get_clean_factor_and_forward_returns`. Furthermore, while I am initially focusing development on the two factor case for simplicity, I think more generally, users should be able to pass as many factors as desired (up to a reasonable amount). In the future, I think there will be parts of the interaction/correlation tear sheet that deal with combinations of more than two factors. 

Regarding implementation, do you think the factor argument should be allowed to take a DataFrame and/or a list of Series?

> Is there an advantage to merge the two factors data together? If so we could call utils.join_factor_with_factor_and_forward_returns inside tears.create_factors_interaction_tear_sheet so that the users can be unaware of this detail.

Initially, I was just trying to keep the code separate for no real good reason. After some thought, I'm thinking there might be good reason to keep the `join_factor_with_factor_and_forward_returns` function. To add a factor post hoc, the user might not want to rerun `get_clean_factor_and_forward_returns` as it would duplicate calculations and take longer than necessary. 

On the other hand, this function only adds minimal additional functionality. As an alternative, the user could just call `quantize_factor` and join the factor and factor_quantiles to the "factor_data" DataFrame without the need for the `join_factor_with_factor_and_fforward_returns` function. 

What's your opinion? Do you think this additional function just clutters up the code base without really adding anything major functionality?

MichaelJMath:
@luca-s , Thanks for the advice on how to improve my commit messages. I’ll start using the prefixes, and work on commenting on what I achieved while letting the code show what I did.

luca-s:
> On this point, I agree with allowing the user to pass factor1 and factor2 data into get_clean_factor_and_forward_returns. Furthermore, while I am initially focusing development on the two factor case for simplicity, I think more generally, users should be able to pass as many factors as desired (up to a reasonable amount). In the future, I think there will be parts of the interaction/correlation tear sheet that deal with combinations of more than two factors.

I wasn't very good in my explanation, what I meant was that I would like the user to call `get_clean_factor_and_forward_returns` on one factor first and then on another factor (not factor1 and factor2 together) and then pass factor1 and factor2  to tears.create_factors_interaction_tear_sheet.
My idea was to have a prototype like this:

```
def create_factors_interaction_tear_sheet(factor_data1, factor_data2):
    ....
```

Or following your suggestion of handling more than 2 factors,  the function could accept a list of factor_data. Also I agree that allowing more than 2 factors makes sense in some parts of the interaction/correlation tear sheet. 

```
def create_factors_interaction_tear_sheet(factor_data_list):
    ....
```

The idea behind this choice is that `factor_data`  is the "standard" input to  most Alphalens functions and I would like to keep it that way to make the new API more user friendly. Internally, create_factors_interaction_tear_sheet can combine the two factors as required to perform the computation, but from the user point of view the new API would follow the current convention.

The issue #253 should be considered too, as it might also come in handy.

Please let me know if you foresee any shortcomings if we go through this path.

MichaelJMath:
> The idea behind this choice is that factor_data is the "standard" input to most Alphalens functions and I would like to keep it that way to make the new API more user friendly. Internally, create_factors_interaction_tear_sheet can combine the two factors as required to perform the computation, but from the user point of view the new API would follow the current convention.
> 
> The issue #253 should be considered too, as it might also come in handy.
> 
> Please let me know if you foresee any shortcomings if we go through this path.

I follow what you are saying now. There aren't any shortcomings that I foresee. Issue #253 addresses my concern about users having to duplicate the forward returns calculation on each call to `get_clean_factor_and_forward_returns`. I will follow that thread, and as that issue gets finalized, I will submit another pull request with changes.

ssanderson:
> The idea behind this choice is that factor_data is the "standard" input to most Alphalens functions and I would like to keep it that way to make the new API more user friendly. Internally, create_factors_interaction_tear_sheet can combine the two factors as required to perform the computation, but from the user point of view the new API would follow the current convention.

@luca-s one worry I have about standardizing on this representation  is that it means duplicating a fair amount of data if you want to analyze multiple factors at the same time.  To take an example from the (awesome by the way!) docstring of `get_clean_factor_and_forward_returns`, the current standard unit of analysis in alphalens looks like this:

```
           -------------------------------------------------------------------
                      |       | 1D  | 5D  | 10D  |factor|group|factor_quantile
           -------------------------------------------------------------------
               date   | asset |     |     |      |      |     |
           -------------------------------------------------------------------
                      | AAPL  | 0.09|-0.01|-0.079|  0.5 |  G1 |      3
                      --------------------------------------------------------
                      | BA    | 0.02| 0.06| 0.020| -1.1 |  G2 |      5
                      --------------------------------------------------------
           2014-01-01 | CMG   | 0.03| 0.09| 0.036|  1.7 |  G2 |      1
                      --------------------------------------------------------
                      | DAL   |-0.02|-0.06|-0.029| -0.1 |  G3 |      5
                      --------------------------------------------------------
                      | LULU  |-0.03| 0.05|-0.009|  2.7 |  G1 |      2
                      --------------------------------------------------------
```

Of the six columns in that frame, only two of them (`factor` and `factor_quantile`) are specific to the factor under analysis.  If a user wants to work with multiple factors, every `factor_data` will duplicate the forward returns and group columns.

luca-s:
@ssanderson I understand your point but I can already see that part of the factor interaction analysis will use the forward returns columns, so I'd keep them. If we end up not useing the forward returns columns then I agree the function should accept a subset of factor_data.

luca-s:
EDIT: deleted my previous comment.
@ssanderson I agree and you point is exactly as @MichaelJMath suggested so let's go with that.

luca-s:
@MichaelJMath @ssanderson by the way, do we all agree we still need the function `utils.join_factor_with_factor_and_fforward_returns` which combine multiple factors with forward returns
 (which will probably make use of the new API developed in #253  )?

PR #269, opened 2018-02-15
Added LICENSE to manifest

alen12345:
We should include the LICENSE in the source code; we also need it to deploy on conda-forge.


twiecki:
@alen12345 We released a new version, can you update conda-forge?

PR #274, opened 2018-02-19
Plotting style

MichaelJMath:
Update plotting style to be consistent with other alphalens tears.

1. Use coolwarm colormap
2. Use plotting.customize to decorate tears function.


luca-s:
thanks :)

PR #275, opened 2018-02-20
TST: added tests for utils.test_get_clean_factor_and_forward_returns

luca-s:
since we are refactoring  `utils.test_get_clean_factor_and_forward_returns `it is good to have non-regression tests before applying any change

### DIFF
--- a/alphalens/utils.py
+++ b/alphalens/utils.py
@@ -235,7 +235,7 @@ def compute_forward_returns(factor_idx,
         # in between the trades so we'll test several entries to find out the
         # correct period length
         #
-        entries_to_test = min(10, len(fwdret.index)-period)
+        entries_to_test = min(10, len(fwdret.index), len(prices.index)-period)
         days_diffs = []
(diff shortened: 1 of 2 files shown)


alphalens/utils.py
```diff
@@ -235,7 +235,7 @@ def compute_forward_returns(factor_idx,
         # in between the trades so we'll test several entries to find out the
         # correct period length
         #
-        entries_to_test = min(10, len(fwdret.index)-period)
+        entries_to_test = min(10, len(fwdret.index), len(prices.index)-period)
```
twiecki: was this a bug?

luca-s: That's a check to avoid an "index out of range" exception but the check was too restrictive and the function didn't work on a small amount of data.So I fixed the check

luca-s:
@twiecki may I merge this or do you see any issue?

twiecki:
Looks great!

Issue #282, opened 2018-02-27
Unhelpful error message if prices and signals do not match


twiecki:
When passing in signals and prices that either don't have the same convention in terms of datetimes or symbol-names this is the error message:

```
/opt/virtualenv/python2/local/lib/python2.7/site-packages/alphalens/utils.pyc in get_clean_factor_and_forward_returns(factor, prices, groupby, binning_by_group, quantiles, bins, periods, filter_zscore, groupby_labels, max_loss)
    533 
    534     merged_data = compute_forward_returns(factor_dateindex, prices, periods,
--> 535                                           filter_zscore)
    536     merged_data['factor'] = factor
    537 

/opt/virtualenv/python2/local/lib/python2.7/site-packages/alphalens/utils.pyc in compute_forward_returns(factor_idx, prices, periods, filter_zscore)
    245             days_diffs.append(period_len.components.days)
    246 
--> 247         delta_days = period_len.components.days - mode(days_diffs).mode[0]
    248         period_len -= pd.Timedelta(days=delta_days)
    249 

UnboundLocalError: local variable 'period_len' referenced before assignment
```

Should be helpful to detect that no signals and prices can be matched and display a clear error message.

PR #283, opened 2018-02-28
2 small fixes

luca-s:
- ENH: use same quantiles colours in different plots 
Cumumlative Returns by Quantile  plot uses a different colour scheme for quantiles than Average Cumulative Returns by Quantile plot

- ENH: improve error message if prices and signals do not match
Issue #282

### DIFF
--- a/alphalens/utils.py
+++ b/alphalens/utils.py
@@ -211,6 +211,11 @@ def compute_forward_returns(factor_idx,
 
     factor_idx = factor_idx.intersection(prices.index)
 
+    if len(factor_idx) == 0:
+        raise ValueError("Factor and prices indices don't match: make sure "
+                         "they have the same convention in terms of datetimes "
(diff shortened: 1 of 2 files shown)


alphalens/utils.py
```diff
@@ -211,6 +211,11 @@ def compute_forward_returns(factor_idx,
 
     factor_idx = factor_idx.intersection(prices.index)
 
+    if len(factor_idx) <= 0:
```
twiecki: `==` :)

alphalens/utils.py
```diff
@@ -211,6 +211,11 @@ def compute_forward_returns(factor_idx,
...
+        raise ValueError("Factor and prices indices don't match: they dont't "
```
twiecki: `dont't` -> `don't`

`Make sure they have same convention...`

PR #287, opened 2018-03-16
BUG: get_clean_factor crashes with malformed 'groupby' data 

luca-s:
If 'groupby' data contains too many or missing values (grouby index is not the same as the factor index) then 'utils.get_clean_factor' and 'utils.get_clean_factor_and_forward_return' crash with meaningless error messages.

+ improved docstring for 'price' argument

### DIFF
--- a/alphalens/performance.py
+++ b/alphalens/performance.py
@@ -966,6 +966,9 @@ def average_cumulative_return_by_quantile(factor_data,
             #
             avgcumret = g_fq.groupby(g_fq).apply(average_cumulative_return,
                                                  demean_by)
+            if len(avgcumret) == 0:
+                continue
+
             avgcumret['group'] = group
             avgcumret.set_index('group', append=True, inplace=True)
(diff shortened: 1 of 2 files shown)


alphalens/performance.py
```diff
@@ -966,6 +966,9 @@ def average_cumulative_return(q_fact, demean_by):
             #
             avgcumret = g_fq.groupby(g_fq).apply(average_cumulative_return,
                                                  demean_by)
+            if len(avgcumret) == 0:
```
twiecki: Is this the fix for the weird crash?

luca-s: No, that's another small bug fix for a very corner case scenario that I discovered while debugging the original bug :) But each fix is in a separate commit, so the git history is clean. Sorry for including everything in this PR but the fixes are so small that I didn't bother to create separate PRs for each small issue.

luca-s:
By the way, the bug was found in here https://www.quantopian.com/posts/alphalens-error

PR #303, opened 2018-05-11
Keep most recent data

jzhang18:
Modified the way to drop nan values for `merged_data`. By this way, it only drops the data that does not have valid `factor value` or `factor_quantile`.


luca-s:
Seems good to me. Just to be sure, Do you need this to avoid dropping data when you have some forward return columns contain valid data but few of them have nan?

jzhang18:
@luca-s yes, that's the reason.

luca-s:
Ok then but I would also add another dropna that discards all the rows for which we have nan in **all** the forward return columns. This would make clear the user didn't pass data for a particular time period. What do you think about this?

twiecki:
After talking about it with @jzhang18 we don't think this should be fixed in alphalens.

PR #304, opened 2018-05-14
WIP First stab at zero-aware quantiles.

twiecki:

PR #306, opened 2018-05-16
ENH: zero aware quantiles

### LINKED ISSUE #281: Add option for "zero aware" quantiles/bins
Sometimes a user might want to analyze a factor where positive and negative values are treated separately. Since the amount of positive values might be different than the negative ones (the factor is biased) the `quantiles` option wouldn't do a good job in separating positive from negative values. Currently the only way to separate positive from negative values is to use the `bins` option, providing custom bin ranges. Is this enough or do we want to improve it?

eigenfoo:
Closes #304 and #281.

This adds the functionality to compute quantile buckets separately for positive and negative signal values. This is useful if your signal/factor is centered and zero is the separation between long and short signals, respectively.

A separate parameter, called `zero_aware`, has been added to the `quantize_factor` function. If True, alphalens will compute the quantiles while being aware of the positive-negative split.


luca-s:
@twiecki don't you want the `zero_aware` flag to be exposed to `get_clean_factor_and_forward_returns`? 

@eigenfoo one suggestion. Since it is almost no effort, why don't we add the `zero_aware` logic to the `bins` option too? As a user I would expect the flag to work with both `bins` and `quantiles`

eigenfoo:
All good suggestions! I'll take care of those.

@twiecki can we confirm that we want the symmetric bucketing: currently, zero is allowed pegged to the "halfway quantile" (e.g. the 50th percentile). I like the idea of symmetric bucketing but that means we'd need to have negative quantile numbers to distinguish between positive and negative factor values. 

Also, I'm concerned about the failing tests: I suspect it has something to do with changes to `pd.qcut` between versions 0.19 and 0.20, although I can't really find anything. Any thoughts @luca-s?

eigenfoo:
@luca-s I don't think I follow: if the quantile numbers don't reflect the positive/negative value of the factor, then how will we distinguish between, say, an asset in the 20th positive quantile and the 20th negative quantile?

I also realize that the `zero_aware` functionality does not make sense with custom quantile/bin ranges. I will add a test to make sure that `zero_aware` is ignored unless `quantiles` or `bins` is an integer. Does that sound reasonable?

luca-s:
@eigenfoo perfect, you can make a check at the beginning of the function and raise a `ValueError` if `zero_aware` is true and `quantiles` or `bins` is not an integer. In this way we can forget about symmetric bucketing and keep the current logic

eigenfoo:
@luca-s  All points have been resolved, except for the tests. It cannot be what you described (i.e. that either the positive or negative subset is empty), because all tests with pandas version >= 0.20 passed successfully.

Perhaps it is to do with [this bug](https://github.com/pandas-dev/pandas/issues/15431) that was patched in pandas 0.20?

luca-s:
@eigenfoo before your latest change the bug was what I said, now that you changed the tests the bug is the one you linked.   Add more data in the biased_factor and you should be able to avoid hitting the bug

eigenfoo:
What an interesting bug; nice catch @luca-s! Ready for review now.

luca-s:
It seems all good now, thanks @eigenfoo . I will go on and merge this

Issue #309, opened 2018-07-23
Biased Mean Quantile Returns for Non-Equal Bins


MichaelJMath:
I was doing a study today, and I came across a potential issue with the `performance.mean_return_by_quantile` logic in the returns tear sheet. I guess it is not so much an issue with the logic of that function, but more so how it is used in building the tear sheet. I think this would only be an issue if a user created bins or quantiles with a non-uniform number of observations in each bin over time. In other words, the amount of stocks in each bin changes over time.

If the number of stocks in a bin is correlated with market direction, this could bias the mean return calculation. For instance, if a factor quantile tends to have a lot of stocks when the market is going up, it will receive a higher weight on those up-days and will thus bias the mean return upwards for that factor quantiles. 

[Here is an example I created.](https://docs.google.com/spreadsheets/d/1vVncTm_TKilb8bYjz5A2BdZ5pkacef-JgJrMApdEzig/edit?usp=sharingl) Note how it affects the mean return of the spread of going long the best quantile and short the worst quantile.

I believe the `mean_return_by_quantile` function already has the ability to group by date, so therefore, you could calculate the mean return for each quantile and date first, and then take the mean by quantile. 

Am I thinking about this right?

luca-s:
Good, point. I totally agree with you. To tell the truth I thought Alphalens was already working accordingly with the solution you are proposing, but unfortunately it is not. 

> I believe the mean_return_by_quantile function already has the ability to group by date, so therefore, you could calculate the mean return for each quantile and date first, and then take the mean by quantile.

I agree and I would do that inside `mean_return_by_quantile`. Currently the function compute the mean returns by date only if `by_date` option is true. We should do that all the time instead and finally take the mean by quantile averaging the daily mean returns only if `by_date` is false.

Do you agree?

MichaelJMath:
I would think that would be the way to go unless there are somehow other use cases where you would actually want the current functionality. However, I can't really think of any cases where that would be the case off the top of my head. So yah, I agree with you.

luca-s:
@MichaelJMath would you like to submit a PR? I tried working on this but I am busy on something else at the moment.

MichaelJMath:
Sure, I'll try to get to it this weekend.

PR #313, opened 2018-08-06
BUG: Calculate mean returns by date before mean return by quantile

MichaelJMath:
Compute mean return by date. If by_date flag is false, then compute
and return the average of the daily mean returns (and also the
standard error of these daily mean returns).

Resolves: Issue #309

### DIFF
--- a/alphalens/performance.py
+++ b/alphalens/performance.py
@@ -650,7 +650,10 @@ def mean_return_by_quantile(factor_data,
     mean_ret = group_stats.T.xs('mean', level=1).T
 
     if not by_date:
-        group_stats = mean_ret.groupby(level='factor_quantile')\
+        grouper = [mean_ret.index.get_level_values('factor_quantile')]
+        if by_group:
+            grouper.append(mean_ret.index.get_level_values('group'))
+        group_stats = mean_ret.groupby(grouper)\
             .agg(['mean', 'std', 'count'])
         mean_ret = group_stats.T.xs('mean', level=1).T
 
--- a/alphalens/performance.py
+++ b/alphalens/performance.py
@@ -638,9 +638,7 @@ def mean_return_by_quantile(factor_data,
     else:
         factor_data = factor_data.copy()
 
-    grouper = ['factor_quantile']
-    if by_date:
-        grouper.append(factor_data.index.get_level_values('date'))
+    grouper = ['factor_quantile', factor_data.index.get_level_values('date')]
 
     if by_group:
         grouper.append('group')
@@ -650,6 +648,11 @@ def mean_return_by_quantile(factor_data,
         .agg(['mean', 'std', 'count'])
 
     mean_ret = group_stats.T.xs('mean', level=1).T
+
+    if not by_date:
+        group_stats = mean_ret.groupby(level='factor_quantile')\
+            .agg(['mean', 'std', 'count'])
+        mean_ret = group_stats.T.xs('mean', level=1).T
 
     std_error_ret = group_stats.T.xs('std', level=1).T \
         / np.sqrt(group_stats.T.xs('count', level=1).T)


MichaelJMath:
It looks like one of the issues is that in older versions of pandas, you can't groupby an index level by specifying the name of the level. I'll look into it a bit later.

MichaelJMath:
The remaining failed unittests don't seem to come from any of the code I edited. They seem to result from a problem in commit e444e8da74a524ce42b172f9931c62597508a749. My logic is as follows:

Below is one of the failed test cases:

```
ERROR: test_average_cumulative_return_by_quantile_0 (alphalens.tests.test_performance.PerformanceTestCase)
----------------------------------------------------------------------
  File "/home/travis/miniconda/envs/testenv/lib/python2.7/site-packages/parameterized/parameterized.py", line 392, in standalone_func
    return func(*(a + p.args), **p.kwargs)
  File "/home/travis/build/quantopian/alphalens/alphalens/tests/test_performance.py", line 920, in test_average_cumulative_return_by_quantile
    0, after + 1), filter_zscore=False)
  File "/home/travis/build/quantopian/alphalens/alphalens/utils.py", line 754, in get_clean_factor_and_forward_returns
    filter_zscore)
  File "/home/travis/build/quantopian/alphalens/alphalens/utils.py", line 304, in compute_forward_returns
    period_len = diff_custom_calendar_timedeltas(start, end, freq)
  File "/home/travis/build/quantopian/alphalens/alphalens/utils.py", line 912, in diff_custom_calendar_timedeltas
    actual_days = np.busday_count(start, end, weekmask, holidays)
TypeError: Iterator operand 0 dtype could not be cast from dtype('<M8[us]') to dtype('<M8[D]') according to the rule 'safe'
```

It seems there is an issue in the way pandas stores dates. (https://stackoverflow.com/questions/31917964/python-numpy-cannot-convert-datetime64ns-to-datetime64d-to-use-with-numba)


I tested this using the following virtual environment configs (matching what was in the unittest):
```
conda create -q -n testenv --yes python=2.7 ipython flake8 numpy scipy nose matplotlib pandas=0.20.3 statsmodels seaborn
```

I tried checking out prior commits and got the error when testing on every commit going back to the e444e8da74a524ce42b172f9931c62597508a749 commit. The commit before that did not generate the error.

This might need to be listed as a separate issue, but I'm not sure why it would just be showing up now on the unittests, and not on prior commits.

Please let me know if I seem to be off track here.

twiecki:
Thanks for tracking this done @MichaelJMath, I bet you are right. Want to do another PR with your fix?

luca-s:
@MichaelJMath I believe the issue has arisen because we don't enforce a particular version of each module Alphalens uses. Currently we only force few pandas and python versions but the other modules might change with time, when new module versions are released. This issue might be due to a new numpy version then.

luca-s:
And the previous comment makes me think we might want to test few specific numpy versions too, as we do we pandas. So that we are sure Alphalens works fine with a specific subset of numpy/pandas/python versions, currently we are only sure about pandas/python versions.

twiecki:
@luca-s Sounds like a good idea, as long as we also test with the most recent version to detect regressions like these.

luca-s:
@twiecki I checked both pyfolio and zipline and they do not force a particular numpy version, so let's keep it this way. As you said, in this way we are able to detect regressions in more recent version of numpy.

MichaelJMath:
@twiecki, I'll work on submitting another pull request this weekend.

twiecki:
@MichaelJMath Appreciate it!

MichaelJMath:
I'll give it a shot. I probably won't get to it this weekend, but I'll put it on my to-do list for next weekend.

luca-s:
@twiecki we can still merge this PR, can't we?

twiecki:
Thanks guys!

luca-s:
@MichaelJMath This weekend I have some time to work on the missing test, so don't worry, I will do it.

PR #317, opened 2018-08-19
BUG: Fix date conversion in newer versions of pandas

MichaelJMath:
In tests using pandas version 20.3, there was an error
`TypeError: Iterator operand 0 dtype could not be cast from
dtype('<M8[us]') to dtype('<M8[D]') according to the rule 'safe'`

Resolves Issue Discovered in #313


luca-s:
Thank you very much @MichaelJMath

Issue #323, opened 2018-10-30
factor_rank_autocorrelation infers turnover period in calendar space while periods could have different time space


twiecki:
https://github.com/quantopian/alphalens/blob/master/alphalens/performance.py#L796 infers the factor frequency by finding the space between factors. It uses pandas and thus returns a time-difference in calendar space. However, `period='10D'` most likely refers to 10 trading days (14 days) if factors and prices are only present for trading days. This then causes a problem in the next line https://github.com/quantopian/alphalens/blob/master/alphalens/performance.py#L797 where `shift=0` and autocorrelation is shown as 1 constantly.

luca-s:
Closed by #324

PR #324, opened 2018-11-03
BUG: factor_rank_autocorrelation infers turnover period in calendar...

luca-s:
…space while periods could have different time space

quantile_turnover is affected by the same bug

closes Issue #323


twiecki:
So this removes the ability to pass in anything more fine-grained than days? That's quite a hefty price-tag for fixing that edge case.

luca-s:
No, don't worry. I didn't change the Alphalens behaviour. Due to the code change I noticed the are few test cases, which I created some time ago, that fail but also they don't make sense. Since it took some time to figure that out, I made sure the functions in `utils` reise now an exception when they receive a wrong parameter type.

luca-s:
Just to be more detailed, the three tests were wrong for the following reason:
- `freq` represents the factor trading calendar, which can be only `Day, BusinessDay or CustomBusinessDay`
- the factor data itself can have any frequency, even hours, minutes, seconds and the `period`we analyse the factor for can use these frequencies even though the trading calendar has daily resolution
- the tests that I removed used the factor data frequency (10min, 1 hour and 3 hour) as value for `freq`, which is wrong
So I deleted the wrong tests but we can still analyse a factor at any frequency we like.

PR #326, opened 2018-12-05
ENH Speed up fwdret computation

twiecki:
By computing directly from prices rather than using `pct_change()`.

### DIFF
--- a/alphalens/utils.py
+++ b/alphalens/utils.py
@@ -278,11 +278,8 @@ def compute_forward_returns(factor,
         #
         # build forward returns
         #
-        fwdret = (prices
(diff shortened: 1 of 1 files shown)


twiecki:
Seeing a 4x increase in speed.

luca-s:
Have you tried multiple version of pandas?

twiecki:
Travis covers those.

luca-s:
I meant if you noticed a speed improvements on multiple pandas version or just one specific version (Quantopian's version I suppose)

twiecki:
I see, I haven't tested that, but I'd be surprised if there were big differences between versions there.

PR #332, opened 2019-01-12
DOC: add GitHub issue template

eigenfoo:
This should make it easier to receive and triage bug reports.

I based the issue template off of [PyMC3's](https://github.com/pymc-devs/pymc3/blob/master/.github/ISSUE_TEMPLATE.md).


luca-s:
Nice! I would also ask for Pandas version

PR #349, opened 2019-09-06
Advanced tutorial on Alphalens (on Quantopian)

luca-s:
@twiecki looking at the issues created on the project I could see that there are several aspects of Alphalens that are misunderstood from the users. This is because the full possibilities of the libraries are not completely covered in the existing tutorials. Since long time ago you asked me if I could add this tutorial to the project here is a PR. The only drawback is that it focuses  on Quantopian API, but at the same time this is useful to give a complete understanding of Alphalens. I'll let you decide if you like to add it to the project.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -49,7 +49,7 @@ Learn more
 ----------
 
 Check out the `example notebooks <https://github.com/quantopian/alphalens/tree/master/alphalens/examples>`__ for more on how to read and use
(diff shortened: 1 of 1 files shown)


twiecki:
@luca-s Great, will take a look soon!

twiecki:
Thanks @luca-s, that reads really well.

PR #360, opened 2020-02-18
WIP: Trying to fix Travis Failures

dmichalowicz:
There's two issues that I see: one is that the latest version of statsmodels requires at least pandas 19, so our pandas 18 builds fail; second is that pandas 18 seems to no longer be on the main [anaconda channel](https://repo.anaconda.com/pkgs/main/linux-64/), so add [conda-forge](https://anaconda.org/conda-forge/pandas/files?sort=basename&sort_order=asc) to the channel list.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,8 @@ before_install:
   - bash miniconda.sh -b -p $HOME/miniconda
   - export PATH="$HOME/miniconda/bin:$PATH"
   - hash -r
-  - conda config --set always_yes yes --set changeps1 no --append channels conda-forge
+  - conda config --set always_yes yes --set changeps1 no
+  - conda config --append channels conda-forge
   - conda update -q conda
   # Useful for debugging any issues with conda
   - conda info -a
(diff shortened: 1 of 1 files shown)


twiecki:
```
0.10s$ conda config --set always_yes yes --set changeps1 no --append channels conda-forge
usage: conda [-h] [-V] [--debug] command ...
conda: error: unrecognized arguments: --append channels conda-forge
The command "conda config --set always_yes yes --set changeps1 no --append channels conda-forge" failed and exited with 2 during .
```

Weird, that's what https://docs.conda.io/projects/conda/en/latest/commands/config.html proposes. You can also just add it to the conda install command using the `-C` flag.

dmichalowicz:
Thanks Thomas, going to try one more thing first...

dmichalowicz:
As of https://seaborn.pydata.org/whatsnew.html#v0-10-0-january-2020 seaborn will break our builds for python 3.5 or less, and pandas 21 or less (I think), so another issue to contend with...

dmichalowicz:
This is no longer relevant as of https://github.com/quantopian/alphalens/pull/363.

Issue #357, opened 2020-01-31
Request: Simplify cumulative_returns definition.


jmccorriston:
## Description

On Quantopian, Alphalens has become more of a central figure as we have been running challenges where the submissions are made as Alphalens tearsheets. In most of these notebooks, the slowest step of running the notebook from top-to-bottom is generating the full Alphalens tearsheet. Recently, I did some profiling of `create_full_tear_sheet` to see if there were any relatively simple opportunities to speed it up. One component that immediately popped up was `cumulative_returns`, which was taking ~65% of the total time of running a full tearsheet, and > 75% of `create_returns_tear_sheet`.

I took a look at the `cumulative_returns` definition, and I was a little surprised by the complexity. Digging into it a bit, it seems the most of the complexity is a product of supporting `period`s that are smaller (faster?) than the frequency of the provided `returns` data. I'm wondering if it would make sense to simplify the implementation of `cumulative_returns` and either drop support for the case where the period is less than or different from the `returns` data.

**High-Level Suggestion**

Without knowing more about the community that uses Alphalens outside of Quantopian, my first suggestion would be to drop support for the case where the `period` is less than the period of the `returns` data. I was surprised that this case was supported, mostly because I didn't realize that `cumulative_returns` was using interpolation to fill in data points that were required to compute the cumulative return for the specified `period`. Additionally, I think it might be a good idea to leverage the [cumulative returns function in empyrical](https://github.com/quantopian/empyrical/blob/master/empyrical/stats.py#L219) so that results are more likely to line up with other quant finance projects/tools. 

By dropping support for the case where the `period` is less then the period of the `returns` data and by implementing `cumulative_returns` in terms of `cum_returns` in `empyrical`, my expectation is that it will become easier to optimize the function for performance.

If there's still a desire to support computing cumulative returns with interpolated returns data, maybe it could be split into a separate function. I read through the code and I don't think I fully understand the current implementation, but I understand that this might be an important use case to some folks.

**Reproducible example:**
Download link: [al_sample_data.csv](https://s3.amazonaws.com/quantopian_static_assets/public/al_sample_data.csv)
```python
import pandas as pd
import cProfile
import pstats
from alphalens.tears import create_returns_tear_sheet

# Include if running in a Jupyter notebook.
%matplotlib inline

al_inputs = pd.read_csv('al_sample_data.csv', index_col=['date', 'asset'], parse_dates=True)

def run_returns_tear_sheet():
[...]

PR #361, opened 2020-02-18
Perf improvements

jmccorriston:
This PR includes changes that address the performance issues highlighted in https://github.com/quantopian/alphalens/issues/357. This change simplifies the `cumulative_returns` computation, which significantly speeds up the `create_returns_tear_sheet` function (and subsequently, `create_full_tear_sheet`).

The changes definitely need input from other folks including @luca-s and someone from engineering at Quantopian (I'll get someone to take a look), so I added a `do not merge` label. I expect I'll need to make pretty significant changes before we can merge.

The branch also sprawled a bit and made a number of other changes, including a few additions and tweaks to functions in `alphalens.utils` as well as some stylistic changes, and minor functional changes. The description below gives a summary of the changes made in this PR.

## Changes

- Drops support for periods other than `1D` in all cumulative returns computations.
  - Cumulative returns computations now use the [cum_returns definition from empyrical].(https://github.com/quantopian/empyrical/blob/master/empyrical/stats.py#L219). This includes the event study tearsheet, which previously used its own definition of cumulative returns. Note that this change makes `alphalens` depend on `empyrical`.
  - The subportfolio technique has been removed from `alphalens.performance`. At least for now, it has been moved into its own function, `subportfolio_cumulative_returns`, in `alphalens.utils`.
- Drops support for all periods that are not exact multiples of a day (i.e. `3D12h` or `1h`) from all turnover metrics.
  - Since we don't expect it to be common that Alphalens users will have factors that update at a precision level smaller than daily, this PR simplifies the turnover tearsheet such that it only computes turnover metrics for periods that are exact multiples of a day.
- Adds `inf` and `nan` handling to the event study tearsheet.
- Adds `inf` handling to `get_clean_factor_and_forward_returns`.
- Added a flag to `get_clean_factor_and_forward_returns` such that it can take returns as input (it could previously only take daily prices).
- Added `backshift_returns_series ` to `alphalens.utils`. This can be used to convert backward-looking returns into a forward-returns series, which is helpful when you only have access to returns data, instead of adjusted daily pricing.


luca-s:
@jmccorriston good job! I will have a look at this PR in the following days.

alphalens/tears.py
```diff
@@ -693,13 +702,20 @@ def create_event_study_tear_sheet(factor_data,
...
             freq=trading_calendar,
-            ax=gf.next_row()
+            ax=gf.next_row(),
+        )
+
+        plotting.plot_cumulative_returns(
+            factor_returns['1D'],
+            period='1D',
+            freq=trading_calendar,
+            ax=gf.next_row(),
```
luca-s: I would totally remove ` plotting.plot_cumulative_returns` call from `create_event_study_tear_sheet`. It is very likely the user will get wrong results.

alphalens/utils.py
```diff
@@ -343,6 +344,37 @@ def compute_forward_returns(factor,
...
+    assert new_date_labels[0] == 0
+
+    new_index = pd.MultiIndex(
+        levels=[new_dates, sids],
+        labels=[new_date_labels, new_sid_labels],
+        sortorder=1,
+        names=ix.names,
+    )
+
+    return pd.Series(data=new_values, index=new_index)
```
luca-s: Is this function used at all?

alphalens/utils.py
```diff
@@ -964,3 +1017,168 @@ def diff_custom_calendar_timedeltas(start, end, freq):
     timediff = end - start
     delta_days = timediff.components.days - actual_days
     return timediff - pd.Timedelta(days=delta_days)
+
+def subportfolio_cumulative_returns(returns, period, freq=None):
+    """
```
luca-s: This function can be totally dropped in my opinion. Without a tear.* function calling this, I can't imagine users being able to make any good use of this functionality.

alphalens/utils.py
```diff
@@ -853,12 +892,26 @@ def std_conversion(period_std, base_period):
...
+    # If exact day multiples are required in the forward return periods,
+    # drop all other columns (e.g. drop 3D12h).
+    if require_exact_day_multiple:
+        pattern = re.compile(r"^(\d+([D]))+$", re.IGNORECASE)
+        valid_columns = [(pattern.match(col) is not None) for col in columns]
+        
+        if sum(valid_columns) < len(valid_columns):
+            warnings.warn(
+                "Skipping return periods that aren't exact multiples" \
+                + " of days."
```
luca-s: I wonder if this has to be a warning (?). The users will be probably puzzled. Not a big deal anyway, just my 2 cents

alphalens/tears.py
jmccorriston: You're right. I'll remove it.

alphalens/utils.py
jmccorriston: It's not used by any of the tearsheets or plotting methods. It's meant to be used as a utility method to help a user transform data into a format which alphalens expects, similar to `get_clean_factor_and_forward_returns`. In this case, `backshift_returns_series` is helpful if you are working with returns data from a pipeline output (like in the example I shared in https://github.com/quantopian/alphalens/issues/357).

alphalens/utils.py
jmccorriston: Ok, I will remove it.

alphalens/utils.py
jmccorriston: My thought was that if the user provides return periods that aren't expecting exact multiples of a day, they might be surprised if they don't see them in the tearsheets. I can remove it if we don't expect this to be a surprise.

alphalens/utils.py
```diff
@@ -852,12 +893,26 @@ def std_conversion(period_std, base_period):
...
+def get_forward_returns_columns(columns, require_exact_day_multiple=False):
     """
     Utility that detects and returns the columns that are forward returns
     """
-    pattern = re.compile(r"^(\d+([Dhms]|ms|us|ns))+$", re.IGNORECASE)
-    valid_columns = [(pattern.match(col) is not None) for col in columns]
+
+    # If exact day multiples are required in the forward return periods,
+    # drop all other columns (e.g. drop 3D12h).
+    if require_exact_day_multiple:
```
dmichalowicz, line 903: This case probably needs a test; I can add.

alphalens/tears.py
```diff
@@ -191,139 +203,150 @@ def create_returns_tear_sheet(factor_data,
...
-        perf.compute_mean_returns_spread(mean_quant_rateret_bydate,
-                                         factor_data['factor_quantile'].max(),
-                                         factor_data['factor_quantile'].min(),
-                                         std_err=compstd_quant_daily)
+    mean_ret_spread_quant, std_spread_quant = perf.compute_mean_returns_spread(
+        mean_quant_rateret_bydate,
+        factor_data["factor_quantile"].max(),
+        factor_data["factor_quantile"].min(),
+        std_err=compstd_quant_daily,
+    )
```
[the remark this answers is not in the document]
jmccorriston: Mostly stylistic. I made a few changes to the plotting of cumulative returns since cumulative returns are now only being computed for one frequency: https://github.com/quantopian/alphalens/pull/361/files#diff-5d20acb213bbb2b2824b8cf64b09e0beR279

alphalens/utils.py
```diff
@@ -639,7 +674,8 @@ def get_clean_factor_and_forward_returns(factor,
                                          groupby_labels=None,
                                          max_loss=0.35,
                                          zero_aware=False,
-                                         cumulative_returns=True):
+                                         cumulative_returns=True,
+                                         is_returns=False):
```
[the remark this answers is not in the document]
jmccorriston: Nice catch on the docstring. I'll add it in. The parameter is used here: https://github.com/quantopian/alphalens/pull/361/files#diff-5195c0fa99041f5ae59720a046520bf0R824. The idea is that the user should be able to provide *either* raw pricing data (which gets converted into forward returns) *or* they can provide forward returns directly. The added flexibility is helpful for the use case where someone is getting returns data from pipeline (instead of pricing data).

alphalens/utils.py
jmccorriston: ty!

alphalens/performance.py
```diff
@@ -329,169 +330,26 @@ def factor_alpha_beta(factor_data,
     return alpha_beta
 
 
-def cumulative_returns(returns, period, freq=None):
```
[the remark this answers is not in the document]
jmccorriston: That was my initial thought. I actually refactored the logic for `period` values other than 1D into a function called subportfolio_cumulative_returns, but luca (the original author) convinced me that we should just remove it: https://github.com/quantopian/alphalens/pull/361#discussion_r383133382, and I agree with him. It will just add confusion. We may want to go back and re-add it later, but if it's not getting used in a tearsheet, I don't think it's worth keeping it in.

alphalens/utils.py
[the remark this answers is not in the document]
luca-s: (sorry for not having caught this earlier) Is `is_return` option really useful? `utils.get_clean_factor` is already available as the entry point when forward returns are available instead of pricing data, that function was made available exactly for this purpose some time ago. I feel like it is more confusing than helpful to have `is_return` option. What do you think?

alphalens/performance.py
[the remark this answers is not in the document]
luca-s: I have just had a last thought on this: why don't we keep the current `cumulative_returns` behaviour and add a special case for the most common scenario that covers 95% of people's use cases? At the beginning of the function we check if we are computing cumulative returns for a daily factor and if so we call `empyrical` otherwise we default to current code. Also we need to plot only 1 day period cumulative returns by default, currently a plot for each period is provided. This would be a less invasive change but still being able to fix the performance issue. On the other hand it would be ugly to compute cumulative returns in two totally different ways depending on the input... I am not sure. What do you think?

setup.py
```diff
@@ -17,6 +17,7 @@
     'seaborn>=0.6.0',
     'statsmodels>=0.6.1',
     'IPython>=3.2.3',
+    'empyrical==0.5.0',
```
twiecki, line 20: I think >= would be better.

alphalens/utils.py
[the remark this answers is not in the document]
jmccorriston: You're right, I think this argument and the subsequent branch in the method should be removed. Instead, the user should use `utils.get_clean_factor`. Thanks for pointing this out, guys!

setup.py
richafrank: 👍

alphalens/performance.py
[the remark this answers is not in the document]
dmichalowicz: I do think that would be a somewhat ugly band-aid type fix. Given the state of this PR and prior discussions I'm inclined to keep Jamie's changes for now, and then we can always reinstate the deleted code in a more carefully organized way if we need it back later.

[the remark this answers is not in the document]
jmccorriston: Luca, that was my initial thought, but it made the code very hard to follow (from my perspective). I think I agree with David that it would make sense to try to work the logic back in later. Is that ok with you?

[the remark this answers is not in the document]
luca-s: Sure, no problem. I just wanted to make sure this possibility has been considered.

alphalens/tests/test_performance.py
```diff
@@ -835,70 +772,75 @@ def test_factor_rank_autocorrelation(self,
...
+             [0.0, 44.1512888], [0.0, 62.0533954], [0.0, 65.8668371],
+             [0.0, 70.4306483]],
+        ),
+        (
+            6, 6, True, False,
+            [[2.0267957, 0.9562173], [2.3876945, 1.3511898], [2.8541303, 1.8856194],  # noqa
+             [3.7228318, 2.6210478], [4.9304831, 3.6296796], [6.6040431, 5.0193734],  # noqa
+             [8.9206837, 6.9404046], [12.127552, 9.6023405], [16.569416, 13.297652],  # noqa
+             [22.727323, 18.434747], [31.272682, 25.584180], [34.358565, 25.497254],  # noqa
+             [37.996459, 25.198051]],
```
dmichalowicz, line 823: It's unfortunate the diff here is so hard to read. This test (`test_common_returns`) and the one below (`test_average_cumulative_return_by_quantile`) had pretty significant changes, but these are expected by @jmccorriston and I according the changes [in this function](https://github.com/quantopian/alphalens/pull/361/files#diff-39e40ed256ddc7619de160850cb6f6a2R642) which uses returns now instead of pricing, and flips the interpretation of the `cumulative` parameter.

dmichalowicz:
The build is now passing for python 2.7 and 3.5, which is the current state of master. I think the last question is in regards to https://github.com/quantopian/alphalens/pull/361/files#r415894312, and if we believe that's right then I think this PR is good.

altquant:
the new call signature to cumulative_returns() will not be compatible with this:
https://github.com/quantopian/alphalens/blob/master/alphalens/performance.py#L933

PR #363, opened 2020-02-25
BLD: Github Actions

gerrymanoim:
This PR transitions us from Travis to Github actions. It also does the work to fix flake8 issues and get tests passing on python 2.7 and python 3.5. 

Tests still fail in python 3.7/3.8 with newer pandas/numpy/etc. 

Future work involves

- Fixing python 3.7/3.8
- Auto tagging for new releases


twiecki:
What does that do?

.github/workflows/main.yml
```diff
@@ -0,0 +1,61 @@
...
+
+jobs:
+  build:
+
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest]
+        python-version: [2.7, 3.5, 3.7, 3.8]
```
richafrank: Is this `python-version` required if we're explicit below in `include`?

alphalens/plotting.py
```diff
@@ -13,20 +13,25 @@
...
+if os.environ.get('DISPLAY', '') == '':
+    print('no display found. Using non-interactive Agg backend')
+    matplotlib.use('Agg')
+import seaborn as sns  # noqa: E402
+import matplotlib.cm as cm  # noqa: E402
+import matplotlib.pyplot as plt  # noqa: E402
+from matplotlib.ticker import ScalarFormatter  # noqa: E402
 
-from functools import wraps
+from functools import wraps  # noqa: E402
```
richafrank: Not a blocker, but maybe for later cleanup. I'd recommend adding a matplotlib rc that uses 'Agg', instead of doing it unconditionally at import time. Then we don't need all the noqa's either.

See https://github.com/quantopian/pyfolio/blob/master/pyfolio/tests/matplotlibrc and https://github.com/quantopian/pyfolio/blob/master/.travis.yml#L36

setup.py
```diff
@@ -24,6 +24,7 @@
         "nose>=1.3.7",
         "parameterized>=0.5.0",
         "tox>=2.3.1",
+        "flake8==3.7.9",
```
richafrank: Should this be a lower bound, since this is the libraries requirements, as opposed to a lockfile?

.github/workflows/main.yml
gerrymanoim: Yes - the include is on the version specified I believe. https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-including-configurations-in-a-matrix-build. My understanding is that you're augmenting and existing build.

PR #372, opened 2020-04-21
Test Fixes + Miscellaneous Cleanup

dmichalowicz:


alphalens/tests/test_performance.py
```diff
@@ -835,70 +772,96 @@ def test_factor_rank_autocorrelation(self,
...
+             [0.0, 0.89993039], [0.0, 1.38620088], [0.0, 2.04972912],
+             [0.0, 2.96841637]],
+        ),
+        (
+            6, 6, True, False,
+            [[0.5410205, 0.0], [0.33326965, 0.0], [0.1770314, 0.0],
+             [0.0670006, 0.0], [4.29107e-4, 0.0], [-0.0228175, 0.0],
+             [0.0000000, 0.0], [0.07500000, 0.0], [0.2125000, 0.0],
+             [0.4282500, 0.0], [0.74522500, 0.0], [1.1961825, 0.0],
+             [1.8272763, 0.0]],
```
dmichalowicz, line 823: These (and the other values above) changed based on https://github.com/quantopian/alphalens/pull/361/files#diff-39e40ed256ddc7619de160850cb6f6a2L848-R689. @jmccorriston and I believe it should in fact change, but these numbers could use a look.

dmichalowicz:
@jmccorriston Ready for a look.

alphalens/tears.py
```diff
@@ -428,12 +429,12 @@ def create_turnover_tear_sheet(factor_data, turnover_periods=None):
...
     if turnover_periods is None:
-        input_periods = utils.get_forward_returns_columns(
+        turnover_periods = utils.get_forward_returns_columns(
             factor_data.columns, require_exact_day_multiple=True,
-        )
-        turnover_periods = list(
-            map((lambda x: pd.Timedelta(x).days), input_periods.get_values())
-        )
+        ).get_values()
+    turnover_periods = list(
```
twiecki, line 435: Is it intentional that you replace the variable? Looks a bit odd.

dmichalowicz: Yes that was intentional, but agree it's confusing to overwrite it twice. I'll switch back to using the helper variable.

